PR #6, opened 2020-05-13
spike with single controller basic functionality

goodviber:
 ModelWizard service to pass individual model/views through registration controller. 
Issue with Formbuilder fieldsets

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -1,0 +1,30 @@
+class RegistrationsController < ApplicationController
+
+  def new
+    byebug
+    step_name = params[:step_name].camelize.constantize
+    @wizard = ModelWizard.new(step_name, session, params).start
+    @registration = @wizard.object
+  end
+
+  def create
+    byebug
+    step_name = params.keys[1]
+    class_name = step_name.capitalize.constantize
+    @wizard = ModelWizard.new(class_name, session, params, params[step_name]).continue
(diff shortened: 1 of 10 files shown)


app/controllers/registrations_controller.rb
```diff
@@ -0,0 +1,30 @@
+class RegistrationsController < ApplicationController
+
+  def new
+    byebug
+    step_name = params[:step_name].camelize.constantize
```
ethax-ross, line 5: I would encapsulate this in the `ModelWizard`

app/services/model_wizard.rb
```diff
@@ -0,0 +1,52 @@
+class ModelWizard
+  attr_reader :object
+
+  def initialize(object, session, params=nil, object_params=nil)
+    @object = object
```
ethax-ross, line 5: Can we call `@object` something else? It gets a bit confusing as well since `@object` starts out as the step name and then becomes an object

app/services/model_wizard.rb
```diff
@@ -0,0 +1,52 @@
...
+    if @object.valid?
+      return process_save
+    end
+    false
+  end
+
+  private
+
+  def load_object
+    class? ? @object.new(@session[@session_params]) : @object
```
ethax-ross, line 38: Not keen on this bit 😬 I think it needs a cleaner/more explicit separation

PR #7, opened 2020-05-14
Factory - single controller with no model wizard

goodviber:
### JIRA ticket number
### Context
Simplified single controller step flow. No need for ModelWizard as each model is single step, simplifying validation process. Each model has next_step method, this could be moved to it's own service, not sure if that would be more scalable. 
Forms now have errors displaying, some issue with fieldsets legends displaying on the Gov Formbuilder


app/models/base.rb
```diff
@@ -0,0 +1,18 @@
+class Base
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  #extend ActiveModel::Naming
```
ethax-ross, line 4: Can we drop the commented out concern?

app/models/base.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def self.step_name
+    name.underscore
+  end
+
+  def step_name
```
ethax-ross, line 10: I can't see value in maintaining this as an instance method as well as a class method?

app/models/identity.rb
```diff
@@ -0,0 +1,13 @@
+class Identity < Base
```
ethax-ross, line 1: It may be clearer to have this match the step name, so `ReturningTeacher`

config/routes.rb
```diff
@@ -5,4 +5,8 @@
   get "/404", to: "errors#not_found", via: :all
   get "/422", to: "errors#unprocessable_entity", via: :all
   get "/500", to: "errors#internal_server_error", via: :all
+
+  get "/registrations/:step_name", to: "registrations#new", as: "new_registration"
+  post "/registrations/:step_name", to: "registrations#create", as: "registrations"
```
ethax-ross, line 10: Can we remove the comment? (or transition these to a `resources` route and drop the two separate routes

spec/models/identity_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'rails_helper'
+
+RSpec.describe Identity do
+  let(:identity) { build(:identity) }
+  let(:no_name) { build(:identity, first_name: "") }
+
+  describe "validation" do
+
+    context "without required attributes" do
+      it "should be invalid" do
```
ethax-ross, line 10: Use of `should` is generally discouraged http://www.betterspecs.org/#should

spec/services/step_factory_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  describe "#create" do
+    context "with a valid constant" do
+      it "instantiates the class" do
+        expect(subject.create("Identity")).to be_an_instance_of(Identity)
+      end
+    end
+
+    context "with an invalid constant" do
+      it "raises a NameError" do
```
ethax-ross, line 14: Might be worth throwing a custom `StepNotFoundError` here, along the lines of:

`Unable to instantiate step: <step_name>`

app/models/base.rb
goodviber: Yep

app/models/base.rb
goodviber: Me neither

app/models/identity.rb
goodviber: Not sure about that, the Model names match the question they are asking, so the next step from Identity is 'are you returning to teaching' - ReturningTeacher

config/routes.rb
goodviber: Yep

spec/services/step_factory_spec.rb
goodviber: Need to do a bit more work here to register permitted classes, as it stand a user could create a String etc.

spec/models/identity_spec.rb
goodviber: old habits...

app/models/identity.rb
ethax-ross: Yep my bad - I misread and thought this step was 'returning teacher'

spec/services/step_factory_spec.rb
goodviber: Have set up a custom error with a redirect to root for now

app/controllers/registrations_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  private
+
+  def update_session_registration_hash
+    session[:registration] ||= {}
+    session[:registration].merge!(registration_params)
+  end
+
+  def entity_name
```
ethax-ross, line 31: Could be replaced with `@registration.step_name.to_sym`

app/models/dob.rb
```diff
@@ -0,0 +1,74 @@
+class Dob < Base
```
ethax-ross, line 1: Can we make this `DateOfBirthStep` for clarity?

app/models/dob.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  def assign_attributes(attributes)
+    new_attributes = attributes.stringify_keys
+    multiparameter_attributes = extract_multiparameter_attributes(new_attributes)
+
+    multiparameter_attributes.each do |multiparameter_attribute, values_hash|
+      set_values = (1..3).collect{ |position| values_hash[position].to_i }
+   
+    if set_values.include?(0)
```
ethax-ross, line 24: Could these not be custom validators and applied next to the presence validator?

app/models/dob.rb
```diff
@@ -0,0 +1,74 @@
...
+      return
+    end
+      
+      self.send("#{multiparameter_attribute}=", Date.new(*set_values))
+    end
+  end
+
+  protected
+
+  def extract_multiparameter_attributes(new_attributes)
```
ethax-ross, line 44: What's this doing? It looks like a very complex way of assigning some attributes from a form?

goodviber: Exactly. The GovFormBuilder sets up 3 attributes for a date form, these are automagically combined when using ActiveRecord, but not with ActiveModel. So I have had to copy the AR method into the model. Its been a bit of a headscratcher TBH.

goodviber: The alternative is to create a custom form with a day, month, year attribute, which seems straightforwards, but it is not clear to me how to create a custom form with error fields using the gov style guide.

app/models/dob.rb
goodviber: See below.

app/models/dob.rb
goodviber: Yep

app/controllers/registrations_controller.rb
goodviber: Good spot

app/models/dob.rb
goodviber: see dob_spec.rb for how I have to build a test model

app/models/dob.rb
goodviber: The errors referenced in the if.. block don't get used, really you can just return from the assign_attributes method, this assigns nil effectively to the dob instance, so the presence check will fail. Not the best solution admittedly.

app/models/uk_completion.rb
```diff
@@ -0,0 +1,13 @@
+class UkCompletion < Base
+  attribute :confirmed, :boolean, default: false
+
+  validates :confirmed, inclusion: { in: [true, false] }
+
+  def next_step
+    if confirmed == true
```
ethax-ross, line 7: Could use a guard here to be more concise - take it or leave it:

```
return "opt_in_emails" if confirmed
nil
```

spec/models/date_of_birth_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'rails_helper'
+
+RSpec.describe DateOfBirth do
+  subject(:date_of_birth) { DateOfBirth.new({'date_of_birth(3i)' => '1', 'date_of_birth(2i)' => '12', 'date_of_birth(1i)' => '2001'}) }
```
ethax-ross, line 4: Since we're not using implicit testing here can we make these `let` instead of `subject`?

spec/models/overseas_completion_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  describe "#next_step" do
+    context "when confirmed is true" do
+      it "returns the correct option" do
+        expect(confirmed.next_step).to eq("opt_in_emails")
+      end
+    end
+
+    context "when unconfirmed" do
+      it "returns nil" do
+        expect(unconfirmed.next_step).to eq(nil)
```
ethax-ross, line 25: `be_nil` - take it or leave it

app/models/date_of_birth.rb
```diff
@@ -0,0 +1,71 @@
...
+      return
+    end
+  
+      self.send("#{multiparameter_attribute}=", Date.new(*set_values))
+    end
+  end
+
+  protected
+
+  def extract_multiparameter_attributes(new_attributes)
```
ethax-ross, line 41: I'm probably missing something here, but does this need to be so complicated? It looks like the input is a hash:

```
{'date_of_birth(3i)' => '1', 'date_of_birth(2i)' => '12', 'date_of_birth(1i)' => '2001'}
```
Why can't we just do:

```
Date.new(hash['date_of_birth(1i)'], hash['date_of_birth(12)'], hash['date_of_birth(3i)'])
```

app/models/uk_completion.rb
goodviber: looks good

spec/models/overseas_completion_spec.rb
goodviber: I'll take it.

app/models/date_of_birth.rb
goodviber: I'm probably missing something myself, I've played around with this for a while, the simple solution would be a custom form with day, month and year params but the gov builder form sends out multiparameter attributes, and this seems to be the standard way of dealing with them in ActiveModel.

ethax-ross: Ok, I don't want to hold up your PR on it - if I get some time I'll take a proper look at wrap my head around it

spec/models/date_of_birth_spec.rb
goodviber: yep

PR #17, opened 2020-05-27
Basic Sitemap

goodviber:
### JIRA ticket number
N/A
Generates a basic sitemap at '/sitemap.xml' .Also provides a rake task for pinging google with updates, this needs to be called when we deploy, and be assigned the root domain.


app/views/sitemaps/index.xml.erb
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+   <url>
+      <loc><%= @host %></loc>
+      <lastmod><%= Time.now.strftime("%Y-%m-%dT%H:%M:%S")%></lastmod>
+   </url>
+  
+   <% @steps.each do |step| %>
+     <url>
+        <loc><%= "#{@host}/registrations/#{step.underscore}" %></loc>
```
jebw, line 10: Can this use the `_url` helpers with host: ... ?, eg

```erb
<%= registrations_path step.underscore, host: @host %>
```

Then it will get updated automatically if the path is changed

lib/tasks/sitemap.rake
```diff
@@ -0,0 +1,9 @@
+desc 'Notify Google of the updated sitemp'
+task :sitemap do
+    require 'net/http'
+    Net::HTTP.get(
+        'www.google.com',
+        '/ping?sitemap=' +
+        URI.escape('http://domain.com/sitemap.xml')
```
jebw, line 7: Is `domain.com` correct or does it need to know the hostname somehow?

app/controllers/sitemaps_controller.rb
```diff
@@ -0,0 +1,11 @@
+class SitemapsController < ApplicationController
+  skip_before_action :http_basic_authenticate # needed for google?
```
jebw, line 2: I don't think this is needed - 2 scenarios I can see

a) its publicly visible, so we want google indexing and this isn't necessary
b) its not publicly visible, so we wont want google to index something thats not accessible?

app/views/sitemaps/index.xml.erb
goodviber: Good point

goodviber: Are you certain about the `skip_before_action`, I have seen other examples which reference it?

lib/tasks/sitemap.rake
goodviber: No, that is just a placeholder, we need to provide either a variable or the actual domain name.

app/views/sitemaps/index.xml.erb
goodviber: Sorry ignore the above comment.

app/controllers/sitemaps_controller.rb
goodviber: If the entire site is publicly visible, we don't need it. If Lizzie want to start using it now, while the site needs authentication, she would need it?

lib/tasks/sitemap.rake
jebw: Can it take it as an ENV VAR in the rake task than, eg

`bundle exec rake sitemap DOMAIN=https://my.domain.com`

goodviber: Could do, or

  URI.escape("#{app.root_url}" + "sitemap.xml")
    )
end
# app.root_url is "http://www.example.com/" for localhost. Can be configured in Application controller
# or environment via default_url_options

goodviber: It would only be needed for the production domain, I would have thought.

PR #23, opened 2020-06-01
Modules

goodviber:
### JIRA ticket number
GITBP-122
### Context
Completed TTA flow as per miro chart
End to end form flow
### Guidance to review
Could possibly remove some of the #validation specs in the child Degree and Equivalent classes as they are inherited.


app/controllers/registrations_controller.rb
```diff
@@ -29,6 +29,8 @@ def update_session_registration_hash
   end
 
   def entity_name
+    # check for namespaces
+    return @registration.step_name.gsub("/","_").to_sym if @registration.step_name.include?("/")
```
ethax-ross, line 33: The conditional isn't necessary:

```
def entity_name
  @registration.step_name.gsub("/", "_").to_sym
end
```

app/helpers/session_helper.rb
```diff
@@ -3,8 +3,8 @@ def show_session(question)
     session[:registration][question] ? session[:registration][question].titleize : nil
   end 
 
-  def show_link(href)
-    "<a href='#{href}'>Change</a>".html_safe
+  def show_link(step)
```
ethax-ross, line 6: This module is missing tests

app/models/degree/secondary_maths_english_grade4.rb
```diff
@@ -0,0 +1,11 @@
+module Degree
+  class SecondaryMathsEnglishGrade4 < SecondaryMathsEnglishGrade4
+    def next_step
+      if has_required_subjects == "yes"
```
ethax-ross, line 4: Why are theses yes/no instead of a boolean value?

app/views/pages/140.html.erb
```diff
@@ -19,40 +19,40 @@
...
+            <p class="govuk-body"><a href='/'>Change</a></p>
         </div>
 
         <div class="govuk-form-group">
             <h3>
                 What subject is your degree in?
             </h3>
-            <p>Archeology</p>
-            <p><a href='/'>Change</a></p>
+            <p class="govuk-body">Archeology</p>
```
ethax-ross, line 46: Might be worth sticking this in a partial/view component to DRY it up

app/controllers/registrations_controller.rb
goodviber: Good point.

app/views/pages/140.html.erb
goodviber: This page is just a prototype, It has to be redone in a completely different format, but good point.

app/models/degree/secondary_maths_english_grade4.rb
goodviber: Have put some basic tests up, the module will probably change following the content review, as the check answers page has to be changed to a GOV format

app/helpers/session_helper.rb
goodviber: Have put some basic tests up, the module will probably change following the content review, as the check answers page has to be changed to a GOV format

app/models/degree/secondary_maths_english_grade4.rb
goodviber: The results are returned as a hash, I thought 'has_required_subjects' => 'yes', is slightly more readable than a bool.

PR #27, opened 2020-06-04
Check answers

goodviber:
Refactoring the check answers pages


app/helpers/session_helper.rb
```diff
@@ -26,4 +26,15 @@ def show_name_and_email
     "#{session[:registration]['first_name'].capitalize}" + " " + "#{session[:registration]['last_name'].capitalize}" + ", " + "#{session[:registration]['email_address']}"
   end
 
+  def show_name
```
ethax-ross, line 29: Can we add tests for these

app/models/primary_or_secondary.rb
```diff
@@ -1,7 +1,7 @@
 class PrimaryOrSecondary < Base
   attribute :primary_or_secondary, :string
 
-  validates :primary_or_secondary, inclusion: { in: %w(primary secondary), message: "Please answer Primary or Secondary." }
+  validates :primary_or_secondary, inclusion: { in: %w(primary secondary), message: "You must select an option" }
```
ethax-ross, line 4: Might be worth making your own inclusion/presence validators so you don't have to repeat the same message everywhere

app/helpers/session_helper.rb
goodviber: Yep

app/models/primary_or_secondary.rb
goodviber: Good idea

.gitignore
```diff
@@ -25,6 +25,8 @@
 /vendor/cache
 yarn-debug.log*
 .yarn-integrity
+#spec state
+/spec/examples.txt
```
ethax-ross, line 29: I thought this was something to do with test parallelisation in Rails 6 🙄

PR #28, opened 2020-06-05
Feature/ui review

DrLazer:
content changes throughout


app/helpers/session_helper.rb
```diff
@@ -1,6 +1,10 @@
 module SessionHelper
   def show_session(question)
-    session[:registration][question] ? session[:registration][question].titleize : nil
+    if question == 'uk_or_overseas' and session[:registration][question] == 'Uk'
```
ethax-ross, line 3: `&&` instead of `and` - I would extract `session[:registration][question]` to a variable as well to DRY this up:

```
some_var = session[:registration][question]
if question == '...' && some_var == 'Uk' ...
```

Also could do with a unit test to cover this

app/models/qualified_to_teach.rb
```diff
@@ -1,10 +1,40 @@
 class QualifiedToTeach < Base
   attribute :qualified_subject, :string
 
-  validates :qualified_subject, inclusion: { in: %w(Maths English Science History) }
+  validates :qualified_subject, inclusion: { in: [
```
ethax-ross, line 4: Is this just for user testing? these values should come from the API dynamically:

```
GET /api/types/dfe_teachingsubjectlist
```

PR #40, opened 2020-06-09
Callback date

goodviber:
Refactor date_of_birth model and add date fields to overseas_candidate model


app/models/date_fields.rb
```diff
@@ -0,0 +1,73 @@
...
+    return if year < start_year || year > end_year
+    
+    begin # catch invalid dates, e.g. 31 Feb
+      self.callback_date = Date.new(year, month, day)
+    rescue ArgumentError
+      return
+    end
+  end
+
+=begin
```
DrLazer: can we can off the commented code?

goodviber: File has been deleted

PR #63, opened 2020-06-16
Bugfix/deploy to test

sfawcett123:
Works with Releases


.github/workflows/build.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - bugfix/deploy-to-test
```
jebw, line 7: This line will need removing before merging

PR #65, opened 2020-06-17
Returners flow

goodviber:
### JIRA ticket number
New returners flow
### Guidance to review
Returning teacher pathway


app/models/previous_id.rb
```diff
@@ -1,7 +1,7 @@
 class PreviousId < Base
   attribute :prev_id, :string
 
-  validates :prev_id, presence: { message: "Please enter your previous teacher ID number" } # is this numeric?
+  #validates :prev_id, presence: { message: "Please enter your previous teacher ID number" } # is this numeric?
```
ethax-ross, line 4: Can we just delete this?

spec/examples.txt
```diff
@@ -1,420 +1,425 @@
 example_id                                                             | status | run_time        |
```
ethax-ross, line 1: Was this not gitignored?

app/models/previous_id.rb
goodviber: I dont have any validation parameters for this input at the moment, might be numerical, might be alpha, although it is optional, it might be used against the CRM, so my thinking is we should do some type of validation once we know what the field looks like, so if they do enter input it must be correct, left this as a reminder.

spec/examples.txt
goodviber: Good point, will get it cached.

PR #67, opened 2020-06-18
Has a degree

goodviber:
### JIRA ticket number
Updates has_a_degree secondary flow


spec/models/start_teacher_training_spec.rb
```diff
@@ -28,4 +28,19 @@
       expect(starter.next_step).to eq("date_of_birth")
     end
   end
+
+  describe "#year_range" do
+    subject { starter.year_range(2) }
+
+    it "returns an array of OpenStructs" do
```
ethax-ross, line 35: Not sure if this test adds value - it's covered in the following test anyway and the fact it's an open struct doesn't really matter?

app/models/start_teacher_training.rb
```diff
@@ -2,6 +2,14 @@ class StartTeacherTraining < Base
   attribute :year_of_entry, :string
 
   validates :year_of_entry, inclusion: { in: %w(2020 2021 2022) }
+  
+  def year_range(number_of_years) # sets year range for view
```
ethax-ross, line 6: Swap the `each` for a `map` and the `to_a` shouldn't be required (I think) then you can drop the `years` initialisation and return the `map` directly:

```
(Date.today.year..Date.today.next_year(number_of_years).year).map do |year|
  years << OpenStruct.new(value: year, name: year)
end
```

spec/models/degree/what_degree_class_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'rails_helper'
+
+RSpec.describe Degree::WhatDegreeClass do
+  let(:what_degree_class) { build(:degree_what_degree_class) }
+  let(:wrong_answer) { build(:degree_what_degree_class, degree_class: "dont know") }
+
+  describe "validation" do
+    context "with invalid subject options" do
+      ['skiing', 'fishing', 'golfing', 'surfing' ].each do |invalid_subject|
```
ethax-ross, line 9: With these I don't think we need to test so many variants -- there's no benefit in testing multiple here

app/models/start_teacher_training.rb
goodviber: Nice. Also on looking at it again had to change the validations, I think we may have to come back to it anyway as it may be covered by an API call?

spec/models/degree/what_degree_class_spec.rb
goodviber: Actually I was thinking as the all the  degree and equivalent specs are covered by their parent class/factory (e.g) we could get rid of all the "validation" specs ?

goodviber: Looking at this again, I pulled this test structure from one of the other DFE apps, it gives a false positive!

PR #69, opened 2020-06-19
Has a degree primary uk completion

goodviber:
### JIRA ticket number
as per title, latest update to flow


spec/models/primary_retake_english_maths.rb
```diff
@@ -0,0 +1,29 @@
+require 'rails_helper'
```
ethax-ross, line 1: Missing `_spec` from the suffix of this file - noticed this in a few other places may be worth doing a quick audit?

goodviber: Yes, am going cross eyed from the number of files...

PR #70, opened 2020-06-22
Changed to blue-green

sfawcett123:
Change from standard deployment to blue_green deployment


terraform/paas/variables.tf
```diff
@@ -23,7 +23,7 @@ variable "additional_routes" {
 
 
 variable "strategy" {
-   default = "standard"
+   default = "blue-green"
```
saliceti, line 26: Isn't it blue-green-v2 ?

PR #71, opened 2020-06-22
Studying for degree

goodviber:
### JIRA ticket number
Studying for a degree and equivalent pathway

### DIFF
--- a/app/models/start_teacher_training.rb
+++ b/app/models/start_teacher_training.rb
@@ -4,9 +4,8 @@ class StartTeacherTraining < Base
   validate :date_cannot_be_in_the_past, unless: :dont_know
   
   def year_range(number_of_years) # sets year range for view
-    years = []
-    (Date.today.year..Date.today.next_year(number_of_years).year).each do |year|
-      years << OpenStruct.new(value: year, name: year)
+    years = (Date.today.year..Date.today.next_year(number_of_years).year).map do |year|
+      OpenStruct.new(value: year, name: year)
     end
     years << OpenStruct.new(value: 'dont know', name: "Don't know")
   end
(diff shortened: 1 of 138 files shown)


app/models/start_teacher_training.rb
```diff
@@ -1,12 +1,14 @@
...
   attribute :year_of_entry, :string
 
-  validate :date_cannot_be_in_the_past
+  validate :date_cannot_be_in_the_past, unless: :dont_know
   
   def year_range(number_of_years) # sets year range for view
-    (Date.today.year..Date.today.next_year(number_of_years).year).map do |year|
-      OpenStruct.new(value: year, name: year)
+    years = []
+    (Date.today.year..Date.today.next_year(number_of_years).year).each do |year|
```
ethax-ross, line 8:
```
years = (Date.today.year..Date.today.next_year(number_of_years).year).map do |year|
  OpenStruct.new(value: year, name: year)
end
years << OpenStruct.new(value: 'dont know', name: "Don't know")
```

app/models/studying/primary_maths_english_grade4.rb
```diff
@@ -0,0 +1,11 @@
+module Studying
+  class PrimaryMathsEnglishGrade4 < PrimaryMathsEnglishGrade4
+    def next_step
+      if has_required_subjects == "yes"
```
ethax-ross, line 4: It looks like we use true/false in some of these and yes/no in others - can they be made consistent?

goodviber: I am converting all to true/false, as I encounter them. Have done this one.

app/models/start_teacher_training.rb
goodviber: 👍

PR #104, opened 2020-07-17
Degree status

goodviber:
### Context
Need to update the degree type and status to match CRM requirement
Do you have a degree?
Yes - status = 222750000 (has degree), type = 222750000 (degree)
No - status = 222750004 (no degree),type = 222750000 (degree)
Studying - status = 222750001 (final)/222750002(second)/222750003(first) (depends on subsequent question), type = 222750000 (degree)
Equivalent - status = 222750000 (has degree), type = 222750005 (equivalent)


app/models/have_a_degree.rb
```diff
@@ -1,18 +1,33 @@
 class HaveADegree < Base
   attribute :degree, :string
+  attribute :degree_type, :string
+
+  before_validation :set_degree_type
 
   validates :degree, types: { method: :get_qualification_degree_status, message: "You must select an option" }
+  validates :degree_type, types: { method: :get_qualification_degree_status }
   # may need updating
   OPTIONS = { yes: "222750000", no: "222750004", studying: "222750001", equivalent: "222750005" }.freeze
```
ethax-ross, line 10: Can we make the studying value here something that's not one of the actual values? It's misleading otherwise as this sets studying to 'final year' - we'd be better setting it to a completely different value, even `studying` so that it's clearer that this won't be written back to the API - maybe add a comment as well to make it clear that it behaves differently?

goodviber: We could do, but then we are validating on the HaveADegree::OPTIONS, rather than the return from the  get_qualification_degree_status Api call. 
`OPTIONS = { yes: "222750000", no: "222750004", studying: "studying", equivalent: "222750005" }.freeze
  DEGREE_TYPE = { degree: "222750000", equivalent: "222750005" }.freeze

  validates :degree, inclusion: { in: OPTIONS.map{ |k,v| v } }
  validates :degree_type, types: { method: :get_qualification_degree_status }`

ethax-ross: If you made it `nil` then you could still validate against the `get_qualification_degree_status` values but add another parameter to allow optional?

```
validates :degree_type, types: { method: :get_qualification_degree_status, presence: false }`
```

Happy if you want to leave it as the final year value, but definitely think it needs a comment to explain why you're doing that and to indicate it will be replaced/overridden in the next step otherwise it will be confusing when people come back to it later on

goodviber: Easy either way

goodviber: Went with "studying", using nil or "" in the messes up the gov formbuilder...

PR #105, opened 2020-07-20
update callback slot to start_at and store class

goodviber:
### Context
We need to map the api keys to the relevant model attributes
updating most Model attributes to match api key
### Guidance to review

Ignore Store class, just currently being used for reference


app/controllers/registrations_controller.rb
```diff
@@ -14,6 +14,7 @@ def create
 
     if @registration.valid?
       update_session_registration_hash
+      byebug
```
ethax-ross, line 17: Byebug

app/models/store.rb
```diff
@@ -0,0 +1,43 @@
+class Store
+  def initialize(session)
+    @session = session
+  end
+
+  def candidate_info
+    x = @session[:registration]
```
ethax-ross, line 7: Is it worth renaming the variables in the individual steps to be consistent with the API? Then you could drop this big mapping function?

ethax-ross: It will also save you mapping the other way around when you come to pre-fill the form with the existing candidate data from the API

goodviber: Yes, if We'd known the names and types could have done it from the get go, plan is to get it working then refactor

app/controllers/registrations_controller.rb
goodviber: swatted

app/models/degree/retake_science.rb
```diff
@@ -1,7 +1,7 @@
 module Degree
   class RetakeScience < RetakeScience
     def next_step
-      if retaking_science == OPTIONS[:yes]
+      if planning_to_retake_cgse_science_id == OPTIONS[:yes]
```
ethax-ross, line 4: Typo in the API here, I'll get this updated

app/models/store.rb
```diff
@@ -0,0 +1,41 @@
...
+      "date_of_birth" => x["date_of_birth"], # DateTime
+      "teacher_id" => x["teacher_id"],
+      "degree_subject" => x["degree_subject"],
+      "telephone" => x["telephone"],
+      "address_line1" => x["address_line1"],
+      "address_line2" => x["address_line2"],
+      "address_city" => x["address_city"],
+      "address_postcode" => x["address_postcode"],
+      "phone_call_scheduled_at" => x["phone_call_scheduled_at"], # DateTime check this
+      "already_subscribed_to_teacher_training_adviser" => x["subscribed"], # Boolean
```
ethax-ross, line 37: You shouldn't pass `already_subscribed_to_teacher_training_adviser` its readonly; you don't need to pass a flag for the subscription, the API will create one if the payload is valid.

Also, can we not drop this big mapping now entirely?

goodviber: The file is just reference, will drop it on the next iteration.

app/models/degree/retake_science.rb
goodviber: ta

goodviber: pushed the update

PR #110, opened 2020-07-23
bump rails version and delete gov formbuilder initializer

goodviber:
### Context
We need to update to latest rails version for the security patch for CVE-2020-8185
update to rails 6.0.3.2, plus general gem update. Gov formbuilder initializer removed as covered by latest version - govuk_design_system_formbuilder 1.2.5 
### Guidance to review
Ensure libffi is up to date. ( `gem install ffi`  if required )


Gemfile
```diff
@@ -4,7 +4,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 ruby File.read(".ruby-version").chomp
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem "rails", "~> 6.0.2"
```
ethax-ross, line 7: I wouldn't drop the pessimistic versioning here - `>=` means you could accidentally bump a major Rails version with `bundle update`

goodviber: Good point.

PR #111, opened 2020-07-24
Registration specs

goodviber:
### Context
We need to ensure the candidate is only allowed to consent to the privacy policy they have viewed, or a policy that exists if they have not viewed any policy.
We need to update the country_id to be set to the UK id value if "UK" is selected, or left as nil if the overseas option is chosen.
Update methods to get and view privacy policies.
Set country_id as secondary attribute on uk_or_overseas model
Some content changes
### Guidance to review
Ignore store.rb, just for reference.


app/models/uk_or_overseas.rb
```diff
@@ -1,7 +1,17 @@
...
 
-  validates :uk_or_overseas, inclusion: { in: %w(UK overseas), message: "Select if you live in the UK or overseas" }
+  before_validation :set_country_id
+
+  validates :uk_or_overseas, inclusion: { in: %w[UK overseas], message: "Select if you live in the UK or overseas" }
+  validates :country_id, types: { method: :get_country_types }, allow_nil: true
+
+  def set_country_id
+    if uk_or_overseas == "UK"
+      self.country_id = ApiClient.get_country_types.find { |v| v.value = "United Kingdom" }.id
```
ethax-ross, line 12: I'd use the `id` here and whack it in a constants file somewhere (whilst not very likely, the name could change to `UK`)

app/views/pages/privacy_policy.html.erb
```diff
@@ -0,0 +1,8 @@
+<!-- Is this rendering correctly? -->
+<div class="content__left">
+  <%= link_to 'Back', 'javascript:history.go(-1);', class: "govuk-back-link hidden", id: "backlink" %>
```
ethax-ross, line 3: Does this not need to work without JS enabled?

app/views/registrations/_accept_privacy_policy.html.erb
```diff
@@ -4,10 +4,11 @@
   <p class="govuk-body">Our privacy notice explains how we use your personal data. It is important that you have read this notice before continuing.</p>
 
   <div class="govuk-form-group">
-    <a href='https://www.gov.uk/help/privacy-notice'>Privacy policy</a>
+  <!-- only permit latest privacy policy acceptance -->
+    <%= link_to "Privacy policy", privacy_policy_path %>
   </div>
 
-  <%= f.govuk_check_box :accepted_policy_id, true, multiple: false, label: { text: "Accept the privacy policy"} %>
+  <%= f.govuk_check_box :accepted_policy_id, ApiClient.get_latest_privacy_policy.id, multiple: false, label: { text: "Accept the privacy policy"} %>
```
ethax-ross, line 11: This should be tied to the policy they are actually accepting - currently if the policy changes whilst they are viewing  the accept policy step they will be marked as accepting a policy we did not link them to. This is how we do it in the app - store the policy id in a hidden field and use the same one to render the link to the policy (the route to view the privacy policy then needs to accept an id to show a specific version, or default to the latest):

https://github.com/DFE-Digital/get-into-teaching-app/blob/master/app/views/event_steps/_further_details.html.erb#L4

https://github.com/DFE-Digital/get-into-teaching-app/blob/master/app/views/event_steps/_further_details.html.erb#L29

spec/factories/accept_privacy_policy.rb
```diff
@@ -1,5 +1,5 @@
 FactoryBot.define do
   factory :accept_privacy_policy do
-    accepted_policy_id { true }
+    accepted_policy_id { ApiClient.get_latest_privacy_policy.id }
```
ethax-ross, line 3: Will this not hit the actual API, or create a VCR file? Probably worth hard coding it in the factories

app/views/pages/privacy_policy.html.erb
goodviber: Yes it is default hidden, the <%= javascript_pack_tag 'back_link'%> displays it

app/views/registrations/_accept_privacy_policy.html.erb
goodviber: I wrote it that way originally, then when I spoke with Myles this morning he said we always want the candidate to agree to just the latest policy. I can revert it after lunch.

app/views/pages/privacy_policy.html.erb
goodviber: I think without JS enabled the pattern is it should not be visible, or  be visible and go back to the previous page, for which I need to implement a previous step method for all models now that the flow is locked down. WIP...

app/views/registrations/_accept_privacy_policy.html.erb
goodviber: I have now put the id into the session, been back and forth with the team over this, I think this is fit for purpose.

spec/factories/accept_privacy_policy.rb
goodviber: Currently it is cached.

app/models/uk_or_overseas.rb
goodviber: Spoke to Ian, probably a good idea, will update at next iteration

spec/factories/accept_privacy_policy.rb
goodviber: Currently it is cached.

app/validators/policy_validator.rb
```diff
@@ -0,0 +1,13 @@
+class PolicyValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    begin
+      policy = ApiClient.send(options[:method], value)
+    rescue GetIntoTeachingApiClient::ApiError => e
```
jebw, line 5: This should probably be removed if there's no clarity on how to handle an error?

Currently this will silently fail which is probably the more problematic outcome. For now I think its better to let the exception propogate, the app will trigger a 500 and if it becomes a specific problem, we can handle it then.

There are 2 likely categories of error
1) Api is down, in which case there's no point progressing things anyway
2) Intermittent network glitches, these should be handled in the api client itself

The client I'd originally written handled retries and my plan is to enhance the new no one with the same behaviour, thats not needed for the beta though.

app/validators/policy_validator.rb
```diff
@@ -0,0 +1,13 @@
...
+      Rails.logger.error e # how do we handle a Bad Request ?
+    end
+
+    unless policy
+      record.errors[attribute] << (options[:message] || "is not included in the list")
```
jebw, line 10: I believe this api is going away in Rails 6.1 ? I think it needs to be `record.errors.add`

app/views/pages/privacy_policy.html.erb
```diff
@@ -0,0 +1,8 @@
...
+  <h2>Privacy Policy</h2>
+  <h3>Legal information</h3>
+  <%= @privacy_policy.text.html_safe %>
+</div>
+<%= javascript_pack_tag 'back_link'%>
```
jebw, line 8: Is there any reason for this to be a separate JS file? Its not a blocker but merging the packs would be one less roundtrip to the server

spec/factories/accept_privacy_policy.rb
jebw: Test suite _really_ shouldn't be hitting the API. I think considering this is a Factory and you're in control of the data its fine to hard code a value. If you do want something random you can use `SecureRandom.uuid` - that is the same structure as the v4 UUIDs used by Dynamics

spec/requests/pages_spec.rb
```diff
@@ -0,0 +1,34 @@
+require "rails_helper"
+
+RSpec.describe PagesController, type: :request do
+  let(:policy) { GetIntoTeachingApiClient::PrivacyPolicy.new(id: "123", text: "Latest privacy policy") }
```
jebw, line 4: I'd suggest replacing `123` with `policy_id` and swap the lines round - it'll be easier to maintain in the future

spec/requests/pages_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  let(:policy_id) { 123 }
```
jebw, line 5: If policy_id is a UUID v4 in CRM, this could be swapped out with `SecureRandom.uuid` ?

spec/requests/pages_spec.rb
goodviber: Good idea

spec/requests/pages_spec.rb
goodviber: Done

spec/requests/pages_spec.rb
goodviber: I think... the factory is built within the test suit, so may use the files in the AcceptPrivacyPolicy cassettes, would have to investigate

PR #123, opened 2020-08-04
Degree status

goodviber:
### Context
we need to match candidate submission data for the different pathways to api validation constraints
1. Candidate Submission service and candidate module
2. Updating form attributes to match api fields
3. Session var for privacy policy
### Guidance to review
Error handling to be returned to when api client updated. 
Linting enforced in script files.


app/controllers/pages_controller.rb
```diff
@@ -7,6 +7,7 @@ def privacy_policy
     policy_id = params[:id]
 
     @privacy_policy = if policy_id
+
```
ethax-ross, line 10: Why do we need to set the session here? Is it not set before we redirect the user anyway? also, random white space...

app/controllers/registrations_controller.rb
```diff
@@ -14,6 +14,7 @@ def create
 
     if @registration.valid?
       update_session_registration_hash
+      candidate_submission if step_name.include?("accept_privacy_policy")
```
ethax-ross, line 17: Can we rename this to make it clearer? so something like `perform_candidate_submission`. Maybe abstract `step_name.include?("...")` to be under a `last_step` method as well, then it would read:

```
perform_candidate_submission if last_step?
```

app/models/equivalent/overseas_completion.rb
```diff
@@ -1,7 +1,7 @@
 module Equivalent
   class OverseasCompletion < OverseasCompletion
     def next_step
-      return "accept_privacy_policy" if confirmed
+      return "equivalent/accept_privacy_policy" if confirmed
```
ethax-ross, line 4: Why do we need a different privacy policy step for each branch?

app/models/have_a_degree.rb
```diff
@@ -1,31 +1,46 @@
...
+  attribute :degree_options, :string
   attribute :degree_status_id, :string
   attribute :degree_type_id, :string
 
+  before_validation :set_degree_status
   before_validation :set_degree_type
 
-  OPTIONS = { yes: "222750000", no: "222750004", studying: "studying", equivalent: "222750005" }.freeze
+  DEGREE_OPTIONS = { yes: "yes", no: "no", studying: "studying", equivalent: "equivalent" }.freeze
+  DEGREE_STATUS_OPTIONS = { yes: "222750000", no: "222750004", studying: "studying" }.freeze
```
ethax-ross, line 10: I think this is going to be confusing when we come back to it later - would it be cleaner if this step only had the `DEGREE_OPTIONS` and the `degree_status_id` and `degree_type_id` were set as part of the submission/where ever you build the payload based on `degree_options`?

app/services/candidate_submission.rb
```diff
@@ -0,0 +1,71 @@
+class CandidateSubmission
+  include Candidate
```
ethax-ross, line 2: Should this be a concern? I don't see it used elsewhere and it's just a bunch of constants - I think the naming and placement is confusing as well, it's not a `Candidate` model

app/services/candidate_submission.rb
```diff
@@ -0,0 +1,71 @@
...
+
+private
+
+  def sign_up_candidate(body)
+    ApiClient.sign_up_teacher_training_adviser_candidate(body)
+  end
+
+  def select_candidate
+    case @step_name
+    when "accept_privacy_policy"
```
ethax-ross, line 21: Determining the type of candidate based on the policy they have accepted feels fickle to me - can we not set an explicit variable in the session with the type of candidate?

app/services/candidate_submission.rb
```diff
@@ -0,0 +1,71 @@
...
+    when "degree/accept_privacy_policy"
+      filter_degree_candidate
+    when "studying/accept_privacy_policy"
+      filter_studying_candidate
+    when "equivalent/accept_privacy_policy"
+      filter_equivalent_candidate
+    end
+  end
+
+  def filter_returner_candidate
```
ethax-ross, line 32: These methods are rinse and repeat; can we DRY it up by passing in the constant, so `DEGREE/RETURNER/etc` to a single method to filter the session based on the candidate type?

app/services/step_factory.rb
```diff
@@ -1,19 +1,81 @@
 class StepFactory
```
ethax-ross, line 1: Does this essentially hold the same information as the `Candidate` model constants but in a slightly different format? if something changes would we have to update both files?

spec/models/candidate_spec.rb
```diff
@@ -0,0 +1,19 @@
+require "rails_helper"
+
+RSpec.describe Candidate do
+  it "has a RETURNER constant" do
```
ethax-ross, line 4: I don't think these specs are adding any value - it should fail elsewhere if these don't exist

spec/services/candidate_submission_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+      "telephone" => "123456",
+      "accepted_policy_id" => ApiClient.get_latest_privacy_policy.id,
+      "phone_call_scheduled_at" => DateTime.now + 1.hour,
+    } }
+  end
+  let(:returner) { described_class.new(session, "accept_privacy_policy") }
+  let(:degree) { described_class.new(session, "degree/accept_privacy_policy") }
+  let(:studying) { described_class.new(session, "studying/accept_privacy_policy") }
+  let(:equivalent) { described_class.new(equivalent_session, "equivalent/accept_privacy_policy") }
+  let(:response_status) { YAML.load_file(VCR.current_cassette.file)["http_interactions"].last["response"]["status"]["code"] }
```
ethax-ross, line 56: I don't think we need to do this - it's just asserting that the static VCR file contains something, we probably just need to ensure that `whatever.call` doesn't raise an exception, which is what would happen if the request changes and no longer matches the VCR cassette:

```
expect(returner.call).to_not raise_error
```

terraform/statuscake/dev.bk.vars
```diff
@@ -1,4 +1,2 @@
 storage_account_name = "s146d01sgtfstate"
-key="tta_statuscake.dev.terraform.tfstate"
```
ethax-ross, line 2: Can we remove the `terraform/` and `tests/` stuff from the diff, it's not changing and it's unrelated to your PR

app/controllers/pages_controller.rb
goodviber: I might be wrong but the first time we offer the privacy policy id from _identity.html.erb, so it the session variable is set from the API call in that view. The second time is the accept_privacy_policy page, so the session variable is set if they have viewed the privacy page from the _identity.html.erb link, if not it is set when they view the privacy page from the accept_privacy_policy link

app/models/equivalent/overseas_completion.rb
goodviber: We did until I realised we can differentiate on the degree_options instead...

app/models/have_a_degree.rb
goodviber: It is quite complex, but we have to put the logic somewhere, I think the attributes should be set in the model, rather than merged later, e.g in the candidate_submission service

app/models/equivalent/overseas_completion.rb
goodviber: we could set another degree_option value in the returning_teacher model, so the value would match the path

app/services/candidate_submission.rb
goodviber: yep, have moved other mods also

app/services/candidate_submission.rb
goodviber: Good idea

app/services/step_factory.rb
goodviber: No, StepFactory is the model classes/ urls, Candidate is the required session values for the api, they are not necessarily dependant.

spec/models/candidate_spec.rb
goodviber: Tell Sonarcloud

terraform/statuscake/dev.bk.vars
goodviber: They had linting "errors" which wouldn't pass the commit until resolved.

spec/services/candidate_submission_spec.rb
goodviber: Good idea.

app/controllers/registrations_controller.rb
goodviber: Will do, although it is refactoring to add more lines 😀

app/controllers/pages_controller.rb
ethax-ross: I don't think the `PagesController` should be concerned/know about the session; it should be set as part of the step that links out to the policy from the wizard -- it's bleeding out into other areas of the app here IMO

app/models/equivalent/overseas_completion.rb
ethax-ross: I think this is disjointed; the type of candidate they are can't naturally be inferred from the privacy policy -- they're the same for each candidate. It should be linked to the logic where that decision is made, so if the candidate is on the 'equivalent' track then the candidate type of 'equivalent' is established when they select that their degree type.

Using the degree type is going to be brittle; we're creating artificial business logic and it's going to be odd/change if the privacy policy is no longer the last step in the journey.

app/models/have_a_degree.rb
ethax-ross: That's fair; can we have some indication that the values that aren't integer ids (so `studying`/`equivalent`) are going to be set in subsequent steps though? even if it's just a comment

spec/models/candidate_spec.rb
ethax-ross: If Sonarcloud is complaining about coverage of these then you're missing test coverage elsewhere

terraform/statuscake/dev.bk.vars
ethax-ross: Urgh, ok must have been merged without CI passing originally - still worth cherry-picking these changes into another PR IMO

app/helpers/session_helper.rb
```diff
@@ -5,7 +5,7 @@ def show_session(question)
   end
 
   def show_link(step)
-    "<a href='#{new_registration_path(step)}'>Change</a>".html_safe
+    "<a href='#{new_registration_path(step)}' style='color:#551A8B'>Change</a>".html_safe
```
ethax-ross, line 8: Can we give this a class instead of hardcoding a style?

app/controllers/pages_controller.rb
goodviber: We can only access the session directly in a view or a controller, not quite sure what you mean.

ethax-ross: Sorry, I’m doing a bad job of explaining this — I’ll grab you tomorrow morning and we can jump on a call 👍

app/helpers/session_helper.rb
goodviber: Sure, no probs, is a temp fix for an odd issue, the text color for the telephone link on the equivalent/overseas completion page was different to the others, not sure why yet.

app/controllers/pages_controller.rb
goodviber: On holiday til Wednesday.

app/models/concerns/candidate.rb
```diff
@@ -0,0 +1,67 @@
+module Candidate
+  RETURNER = %i[
+    first_name
+    last_name
+    date_of_birth
+    address_line1
```
mattonfoot, line 6: this should be `address_line_1` and `address_line_2`

app/models/concerns/candidate.rb
```diff
@@ -0,0 +1,67 @@
...
+    uk_degree_grade_id
+    preferred_education_phase_id
+    has_gcse_maths_and_english_id
+    planning_to_retake_gcse_maths_and_english_id
+    has_gcse_science_id
+    planning_to_retake_gcse_science_id
+    initial_teacher_training_year_id
+    date_of_birth
+    country_id
+    address_line1
```
mattonfoot, line 37: this should be `address_line_1` and `address_line_2`

app/models/equivalent/overseas_completion.rb
goodviber: I have altered the candidate submission to rely on the degree_options. It needs QA and will have to rm the extra accept_privacy_policy models, extend testing for returning_teacher...

app/controllers/pages_controller.rb
ethax-ross: Ah ok, in that case I'll try and write it down so I don't forget:

I believe the way it works on the privacy policy step is that we render a hidden field containing the `accepted_policy_id` and that gets posted back to the API and subsequently set in the session (which is how all the other steps work in general). The thing I don't understand here is that we've now also got similar logic in the `PagesController`... why not be consistent and add a hidden field to the `identity` step as well, so you can remove this from the `PagesController`?

https://github.com/DFE-Digital/get-teacher-training-adviser-service/blob/master/app/models/accept_privacy_policy.rb#L2

https://github.com/DFE-Digital/get-teacher-training-adviser-service/blob/master/app/models/identity.rb#L2-L4

app/helpers/session_helper.rb
ethax-ross: May be worth flagging, could just be an oversight

app/models/have_a_degree.rb
goodviber: We should put it in the READme, or wiki.

app/controllers/pages_controller.rb
goodviber: The pages work differently, the accept_privacy_policy model has :accepted_policy_id as an attribute, and a link, the identity page just has a link. I have been back and forth over it, deleting the session variable between views etc, and as far as I can see it works perfectly. 
Both pages use `<%= link_to "privacy notice", privacy_policy_path(:id => session[:privacy_policy_id].presence || ApiClient.get_latest_privacy_policy.id) %>` . There is the session[:privacy_policy_id] which we use in the views and session[:registration]["accepted_policy_id"] which we send to the api. I see what you mean, however I don't think it should be the responsibility of the Identity model to set the accepted_policy_id attribute.

app/models/equivalent/overseas_completion.rb
goodviber: Done

PR #130, opened 2020-08-13
Degree status updated

goodviber:
### JIRA ticket number
WIP DO NOT APPROVE/MERGE
### Context
we need to match candidate submission data for the different pathways to api validation constraints
Candidate Submission service and candidate module
Updating form attributes to match api fields
Session var for privacy policy
### Guidance to review
Error handling to be returned to when api client updated.
Refactoring of integer attributes


app/helpers/session_helper.rb
```diff
@@ -5,7 +5,7 @@ def show_session(question)
   end
 
   def show_link(step)
-    "<a href='#{new_registration_path(step)}'>Change</a>".html_safe
+    "<a href='#{new_registration_path(step)}' class='link-color'>Change</a>".html_safe
```
ethax-ross, line 8: Did you run this past Ben?

app/models/have_a_degree.rb
```diff
@@ -1,31 +1,46 @@
 class HaveADegree < Base
+  attribute :degree_options, :string
   attribute :degree_status_id, :string
   attribute :degree_type_id, :string
-
+# change this again d
```
ethax-ross, line 5: Typo - is this still needed?

spec/models/concerns/candidate_spec.rb
```diff
@@ -0,0 +1,19 @@
+require "rails_helper"
+
+RSpec.describe Candidate do
+  it "has a RETURNER constant" do
```
ethax-ross, line 4: Were we not dropping these?

spec/requests/pages_spec.rb
```diff
@@ -30,5 +30,24 @@
...
     end
+
+    context "viewing a specific privacy policy" do
+      subject do
+        get privacy_policy_path(id: policy.id)
+        response
+      end
+
+      before do
+        allow_any_instance_of(GetIntoTeachingApiClient::PrivacyPoliciesApi).to receive(:get_privacy_policy).with(policy.id).and_return(policy)
```
ethax-ross, line 41: Can we have a test here to ensure the session is set?

app/services/candidate_submission.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+  end
+
+  def filter_candidate(path)
+    data = @session[:registration].select { |key, _| path.include? key.to_sym }
+    merge_integers(data)
+  end
+
+  def merge_integers(body)
+    str_values = %w[
```
ethax-ross, line 37: I think this should be fixed before this PR is merged down. In my mind it should just be changing these attributes to type `:integer`:

https://github.com/DFE-Digital/get-teacher-training-adviser-service/blob/e74328ddae494c9580ec7d22e7bb4ef37807649e/app/models/what_degree_class.rb#L4

And then `to_s` the value here:

https://github.com/DFE-Digital/get-teacher-training-adviser-service/blob/master/app/validators/types_validator.rb#L5

spec/models/concerns/candidate_spec.rb
goodviber: 👍

spec/requests/pages_spec.rb
goodviber: 👍

app/helpers/session_helper.rb
goodviber: Yes, figured out the issue, put it on another ticket.

app/models/have_a_degree.rb
goodviber: Removed

app/services/candidate_submission.rb
goodviber: Not opposed to doing this, but it needs to be a sep PR, it touches a lot of files, and the verification process to check the correct data is getting posted is prolonged, I have opened a ticket.

jebw:
Is this still relevant or can it be closed?

PR #131, opened 2020-08-14
update overseas telephone link

goodviber:
### JIRA ticket number
https://dfedigital.atlassian.net/browse/GITPB-596
### Context
The link to Telephone on equivalent/overseas_completion page was displaying as "not visited"
updated telephone partial to provide correct equivalent link

### DIFF
--- a/app/views/check_answers/_telephone.html.erb
+++ b/app/views/check_answers/_telephone.html.erb
@@ -6,7 +6,8 @@
     <p class="govuk-body"><%= show_phone %></p>
   </dd>
   <dd class="govuk-summary-list__actions">
-    <% if defined?(link) && link == Equivalent::LINK %>
+    <% if defined?(link) %>
+      <% if link == Equivalent::LINK %>
         <%= show_link("#{link}" + 'overseas_candidate') %>
       <% else %>
         <%= show_link("#{link}" + 'overseas_telephone') %>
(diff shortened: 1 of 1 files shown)


app/views/check_answers/_telephone.html.erb
```diff
@@ -7,8 +7,12 @@
   </dd>
   <dd class="govuk-summary-list__actions">
     <% if defined?(link) %>
```
ethax-ross, line 9: This section could be simplified to:

```
<% if defined?(link) && link == Equivalent::LINK %>
        <%= show_link("#{link}" + 'overseas_candidate') %>
<% else %>
        <%= show_link("#{link}" + 'overseas_telephone') %>
<% end %>
```

Also, one for another PR but it's not very clear what step `overseas_candidate` is - you have to dig up to the parent model:

```
OverseasCandidate < Callback
```

^ this inheritance would be clearer if the model was something like `OverseasCallback` or `OverseasCandidateCallback`

goodviber: Not sure that works, the else clause at line 15 needs to cover the 'no link' case ?

goodviber: image

ethax-ross: Ah you're right, I misread that - thought lines 12 and 15 were the same

PR #132, opened 2020-08-17
Update vcap configuration

jebw:
### Context

The current preprod and rolling environments assume the presence of the VCAP_SERVICES env var - and the production environment does not use the VCAP_SERVICES env var to configure redis.

This prevents testing preprod and rolling environments locally and will also require further configuration of the production environment when it is deployed onto PaaS.

1. Always parse VCAP_SERVICES configuration into the `Rails.application.config` hash if it is set
2. If `REDIS_URL` is not set, and a redis config is present in the VCAP_SERVICES, use that to configure Redis
3. Check for Redis connectivity at boot and fail if not present.

### Guidance to review

This is set to deploy straight through to the rolling environment at present, as if it was on `master`. If that environment is still working than these changes are configured up correctly.

### Merge notes

Last commit is for testing purposes and needs to be dropped before merge to `master`.

### DIFF
--- a/config/environments/preprod.rb
+++ b/config/environments/preprod.rb
@@ -2,16 +2,5 @@
 require File.expand_path("production.rb", __dir__)
 
 Rails.application.configure do
-  # Override any production defaults here
-  # Use a different cache store in preprod.
-  redis_url = JSON.parse(ENV["VCAP_SERVICES"])["redis"][0]["credentials"]["uri"]
-  config.cache_store = :redis_cache_store, {
-    url: redis_url,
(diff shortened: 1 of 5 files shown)


sfawcett123:
Please be sure to remove debug commit before Merge.

Also as a thought, is it possible to add a entry in health check to say if redis is connected or not?

jebw:
Good call adding Redis to the healthcheck - I think its a separate consideration to this though.

I want to take another look at the healthcheck endpoint anyway because I don't think its reflecting the correct git sha so I'll add something when I look at that

PR #134, opened 2020-08-17
Booking quotas

goodviber:
### JIRA ticket number
https://dfedigital.atlassian.net/browse/GITPB-588
### Context
The booking quotas api call returns a list of timeslots which can potentially time out if the candidate takes too long to complete the form
Remove the present days timeslots from the available drop down selection on the equivalent/overseas_candidate and equivalent/uk_callback pages


app/models/callback.rb
```diff
@@ -1,23 +1,28 @@
...
+        data.each do |x|
+          gmt_start_slot = Time.zone.parse(x.start_at.to_s).strftime("%I:%M %P")
+          gmt_end_slot = Time.zone.parse(x.end_at.to_s).strftime("%I:%M %P")
+          options_hash[day] << [gmt_start_slot.to_s + " - " + gmt_end_slot.to_s, x.start_at]
+        end
       end
+      next_day_check(options_hash)
+    end
+
+    def next_day_check(options)
```
ethax-ross, line 23: Can we rename this to `remove_current_day` for clarity? You could also just skip it in the `each` loop as well with something like:

```
next if Date.parse(day) == Date.today
```

app/models/callback.rb
```diff
@@ -1,23 +1,28 @@
...
-      data.each do |x|
-        gmt_start_slot = Time.zone.parse(x.start_at.to_s).strftime("%I:%M %P")
-        gmt_end_slot = Time.zone.parse(x.end_at.to_s).strftime("%I:%M %P")
-        options_hash[day] << [gmt_start_slot.to_s + " - " + gmt_end_slot.to_s, x.start_at]
+  class << self
+    def options
+      quotas = ApiClient.get_callback_booking_quotas
+      grouped_quotas = quotas.group_by(&:day)
+      options_hash = Hash.new { |hash, key| hash[key] = [] }
+      grouped_quotas.each do |day, data|
```
ethax-ross, line 13: This is quite busy/hard to read, would something like this work?

```
def options
  quotas_by_day = ApiClient.get_callback_booking_quotas.group_by(&:day)
  quotas_by_day.transform_values do |quotas|
    quotas.map { |quota| option(quota) }
  end

  today = Time.now.strftime("%A")
  quotas.delete(today)
end

def option(quota)
  start = Time.zone.parse(quota.start_at.to_s).strftime("%I:%M %P")
  end = Time.zone.parse(quota.end_at.to_s).strftime("%I:%M %P")

  options_hash[day] << ["#{start.to_s } - #{end.to_s}", quota.start_at]
end
```

app/models/equivalent/uk_callback.rb
```diff
@@ -1,15 +1,17 @@
...
   class UkCallback < Callback
-    def self.options
-      quotas = ApiClient.get_callback_booking_quotas
-      grouped_quotas = quotas.group_by(&:day)
-      options_hash = Hash.new { |hash, key| hash[key] = [] }
-      grouped_quotas.each do |day, data|
-        data.each do |x|
-          options_hash[day] << [x.time_slot, x.start_at]
+    class << self
+      def options
```
ethax-ross, line 4: Why is this different to the method in the parent class? We seem to convert to GMT in that one but not here, but I don't get why? Also this could be DRYer if we just abstract the method that creates the option array and override that instead of the full method

spec/support/shared_contexts/callback_next_day_check.rb
```diff
@@ -0,0 +1,11 @@
+RSpec.shared_context "callback_next_day_check" do
```
ethax-ross, line 1: Why have 3x shared contexts here instead of just one? It makes it hard to follow what's going on

app/models/callback.rb
goodviber: Happy to rename, I don't think it is within a loop, it takes the result of the ::options method (options_hash) and performs a shift if the first key is today.

app/models/equivalent/uk_callback.rb
goodviber: I think it is different as for uk callbacks we show BST ?

app/models/callback.rb
goodviber: can't quite get that to work, quotas.delete(today) would return today, end is a reserved word, options_hash is not declared, have removed extra grouped quotas line, think it looks ok.

ethax-ross: oh I just have a typo, should be:

```
def options
  quotas_by_day = ApiClient.get_callback_booking_quotas.group_by(&:day)
  quotas_by_day.transform_values do |quotas|
    quotas.map { |quota| option(quota) }
  end

  today = Time.now.strftime("%A")
  quotas_by_day.delete(today)
  quotas_by_day
end
```

app/models/callback.rb
ethax-ross: I mean you could drop the method and put that in the `each` loop, so:

```
grouped_quotas.each do |day, data|
  next if Date.parse(day) == Date.today
  ...
end
```

app/models/equivalent/uk_callback.rb
ethax-ross: Oh ok, so the x.time_slot is in BST already? makes sense - in that case I would just abstract constructing the option to a new method in the superclass and override that here so you don't need to re-write the whole method

spec/support/shared_contexts/callback_next_day_check.rb
goodviber: There are 2 shared contexts and 2 shared examples. To test the options method, we need `let(:options_hash) { described_class.options }`, to test the remove_current_day method, we need `let(:options) do
    { Date.today.strftime("%a %d %B") => [
      ["todays times and dates"], ["some more todays times and dates"]
    ],
      Date.tomorrow.strftime("%a %d %B") => [
        ["tomorrows times and dates"], ["some more tomorrows times and dates"]
      ] }
  end
  let(:options_hash) { described_class.remove_current_day(options) }`

Not really a fan of shared contexts myself, you always have to hop around to find out what's going on...

app/models/equivalent/uk_callback.rb
goodviber: have abstracted the grouped_quotas, and set a test. Am happy with the remove_current_day, not sure of the values of putting a function within a loop just to skip it ? Also it is very readable.

spec/support/shared_contexts/callback_next_day_check.rb
goodviber: Actually, one of those was duplicated, have removed

app/models/callback.rb
goodviber: Yes, sorry I misunderstood

app/models/callback.rb
jebw: Some of the complexity seems to be combining logic for what slots are available, with the presentation of times for those slots? Can the model have a more structured storage of the times (as DateTime objects)? And then use a helper method to map the values on the form in the view partial - eg

```ruby
f.select :my_attribute, convert_time_zones_of_quotas(f.object.options), :first, :last
```

app/models/equivalent/uk_callback.rb
jebw: Can the grouping of options for presentation end up in a helper for use in the view - surely the data model doesn't/shouldn't care how this is getting shown in a select? Just which slot is chosen

app/models/callback.rb
```diff
@@ -1,23 +1,30 @@
...
-      data.each do |x|
-        gmt_start_slot = Time.zone.parse(x.start_at.to_s).strftime("%I:%M %P")
-        gmt_end_slot = Time.zone.parse(x.end_at.to_s).strftime("%I:%M %P")
-        options_hash[day] << [gmt_start_slot.to_s + " - " + gmt_end_slot.to_s, x.start_at]
+  class << self
+    def grouped_quotas
+      ApiClient.get_callback_booking_quotas.group_by(&:day)
+    end
+
+    def options
```
jebw, line 13: This looks like view code and could be moved out to a helper method - you could have an option on the helper method to strip out today, than have your view opt in to it where appropriate - something like the following

```ruby
# my_data_model.rb
class MyDateModel
  def get_dates
    ApiClient.get_callback_booking_quotas 
   # massage this into nice structured data if not already
  end

  def get_dates_exc_today
    get_dates.reject { |d| d.day == Time.zone.today }
end

# booking_dates_helper.rb
def group_booking_dates(dates)
  # logic for grouping dates, and formatting them
  # See Ross' comments
end

# _my_form_partial.rb
f.grouped_collection_select :my_slot_attribute, 
    group_booking_dates(f.object.get_quotas_exc_today), 
    :first, 
    :last,
    :first,
    :last
```

spec/models/callback_spec.rb
```diff
@@ -41,8 +41,32 @@
...
+    end
+  end
+
   describe "#self.options" do
-    let(:options_hash) { Callback.options }
+    include_context "callback_options_hash"
     include_examples "callback_options"
   end
+
+  describe "#self.remove_current_day" do
```
jebw, line 56: If moving the logic out to the view, this will need to move to a helper spec.

jebw (commented):
I think some of the complexity is combining date formatting logic, exclusion of todays slots, and structuring data for a collection select all into a single method.

This then complicates overriding just part of it in a child class.

spec/models/callback_spec.rb
goodviber: 👍

app/models/callback.rb
goodviber: 👍

PR #142, opened 2020-08-18
Type id active model attributes as :integer

ethax-ross:
### JIRA ticket number

[GITPB-602](https://dfedigital.atlassian.net/jira/software/projects/GITPB/boards/61?selectedIssue=GITPB-602)

### Context

The types for most `_id` active model attributes will be integer (they can also be GUIDs for certain types, such as preferred teaching subjects and countries). These are currently typed as `string` (and will come back from the HTML form submissions as a string). If we type the model attributes as `:integer` it means we don't have to convert these values to integers prior to submission to the API.

- Type `id` active model attributes as :integer where possible
- Update `StageInterestedTeaching` model to be consistent with others (generating options from API)


app/models/start_teacher_training.rb
```diff
@@ -1,7 +1,7 @@
 class StartTeacherTraining < Base
   extend ApiOptions
 
-  attribute :initial_teacher_training_year_id, :string
+  attribute :initial_teacher_training_year_id, :integer
```
goodviber, line 4: Looks good, just the inital typo is causing both `"intital_teacher_training_year_id"=>"12917"` and `"initial_teacher_training_year_id"=>12917` to appear in the session

PR #147, opened 2020-08-19
Clean up use of VCR cassettes

ethax-ross:
### JIRA ticket number

[GITPB-603](https://dfedigital.atlassian.net/jira/software/projects/GITPB/boards/61?selectedIssue=GITPB-603)

### Context

Currently we generate a cassette per test, with lots of duplicate cassettes being recorded. Instead, we should generate the cassettes once per unique request and share them between tests.

- Remove current VCR cassettes

- Remove per-test VCR config

- Record VCR cassettes per-request

We used to record VCR cassettes per test, however this resulted in a lot of duplicate cassettes (the same requests in different specs). Instead, this commit records all GET requests automatically and shares them between specs.

- Reduce number of privacy policy cassettes

Testing for an empty policy id is the same as testing for an invalid one as we hit the API in both instances. We use `invalid_id` elsewhere instead of `hello`, so updating reconciles those cassettes.


goodviber (approved):
lgtm

PR #149, opened 2020-08-19
Degree status again

goodviber:
### Context
To enable candidate submission we need to update the degree_status_is and degree_type to match the api constraints 
update returning_teacher models with degree_options attributes for UI and education phase for api constraint
update have_a_degree with degree_options and hooks to set degree_status_id and degree_type


app/models/returning_teacher.rb
```diff
@@ -1,8 +1,23 @@
 class ReturningTeacher < Base
   attribute :returning_to_teaching, :boolean
+  attribute :degree_options, :string
+  attribute :preferred_education_phase_id, :integer
+
+  DEGREE_OPTIONS = { returner: "returner" }.freeze
+
+  before_validation :set_degree_options
```
ethax-ross, line 8: Instead of doing these in a `before_validation` I think you could just default them as they are static:

```
attribute :degree_options, :string, default: -> { DEGREE_OPTIONS[:returner] }
attribute : preferred_education_phase_id, :integer, default: -> { StageInterestedTeaching::OPTIONS[:secondary].to_i }
```

spec/models/have_a_degree_spec.rb
```diff
@@ -48,4 +48,64 @@
...
+    context "with degree_option degree" do
+      it "sets the correct degree_type_id before validation" do
+        have_a_degree.valid?
+        expect(have_a_degree.degree_type_id).to eq(HaveADegree::DEGREE_TYPE[:degree])
+      end
+    end
+
+    context "with degree_option no" do
+      it "sets the correct degree_type_id before validation" do
+        no.valid?
```
ethax-ross, line 92: It may be worth overriding the setter for `degree_options` instead of doing the logic in `before_validation` hooks as it may catch us out in the future by reading the value before running the `.valid?` method

ethax-ross (commented):
Looks good, couple of minor suggestions

app/models/returning_teacher.rb
goodviber: Can do the same with `attribute :preferred_education_phase_id, :integer, default: StageInterestedTeaching::OPTIONS[:secondary].to_i`

spec/models/have_a_degree_spec.rb
goodviber: Sorry, I don't quite follow ?

ethax-ross: So at the minute if you did something like:

```
model.degree_options = "degree"
```

You need to know to call `.valid?` before trying to read the `model.degree_status_id` or `model.degree_type_id`. It feels like something that will potentially trip us up in the future. Instead, you could override the setter of `degree_options`, something like this:

```
def degree_options=(value)
  super
  set_degree_status
  set_degree_type  
end
```

That way you can do this and get the correct values:

```
model.degree_options = "degree"
puts model.degree_status_id
puts model.degree_type_id
```

^ no longer needing the call to `valid?` in between

app/models/returning_teacher.rb
goodviber: default takes a block or a string, heh ho...

spec/models/have_a_degree_spec.rb
goodviber: 👍

spec/models/returning_teacher_spec.rb
```diff
@@ -13,6 +13,20 @@
     end
   end
 
+  describe "#set_degree_options" do
+    it "sets the degree option before validation" do
+      returning_teacher.valid?
```
ethax-ross, line 18: Shouldn't need these `valid?` calls any more as it's set as a default on the attribute

goodviber: 👍

ethax-ross (approved):

PR #151, opened 2020-08-19
rm unused models

goodviber:
### Context
we have some old unused models which need pruning
rm old completion page models


ethax-ross:
Can we remove the associated factories as well? Looks like some may still be kicking around:

https://github.com/DFE-Digital/get-teacher-training-adviser-service/blob/master/spec/factories/uk_new_primary_completion.rb

ethax-ross (approved):

PR #153, opened 2020-08-19
Issue verification code on matchback

ethax-ross:
### JIRA ticket number

[GITPB-415](https://dfedigital.atlassian.net/jira/software/projects/GITPB/boards/61?selectedIssue=GITPB-415)

### Context

When a candidate enters their first/last/email we want to check if we have a matching candidate in the system already. If we do, a verification code should be emailed to the candidate that they can then use to 'authenticate' and we can pre-fill the sign up form.

- Move logic to save into the step model

This will enable us to perform other model-specific actions - such as requesting out to issue a verification code on the `Identity` step - on save.

Also renames `@registration` to `@step` for clarity.

- Move responsibility of updating store to the steps

Previously the `RegistrationsController` updated the store; this is now encapsulated into the `Step` model so that we can more easily structure the authentication/matchback logic (i.e. clear the store prior to pre-populating on successful matchback).

- Send verification code on matchback

If a candidate enters a first/last/email that matches an existing candidate in the CRM we want to email them a verification code that can be used in a subsequent step to retrieve their details and pre-fill the form.

### Guidance to review

The `IssueVerificationCode` concern is pretty much lifted/shifted from the get-into-teaching-app; the other minor changes aim to align this registration process with the app as well.

Screenshot 2020-08-19 at 15 46 14

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -1,7 +1,7 @@
 class RegistrationsController < ApplicationController
   def new
     step_name = params[:step_name]
-    @step = StepFactory.create!(step_name)
+    @step = StepFactory.create!(step_name, store)
   rescue StepFactory::NameNotFoundError => e
     # needs to redirect to root or last valid step
     redirect_to :root, alert: "Error: #{e.message}"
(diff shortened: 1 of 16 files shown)


goodviber (approved):

ethax-ross:
I'm going to raise a new PR with the changes on top of the refactoring PR.

PR #156, opened 2020-08-20
Simplify StepFactory to remove hard coded steps

ethax-ross:

### DIFF
--- a/app/controllers/sitemaps_controller.rb
+++ b/app/controllers/sitemaps_controller.rb
@@ -1,7 +1,7 @@
 class SitemapsController < ApplicationController
   def index
     @host = "#{request.protocol}#{request.host}"
-    @steps = StepFactory::STEP_NAMES
+    @steps = Base.descendants.map(&:to_s)
     respond_to do |format|
       format.xml
     end
--- a/app/services/step_factory.rb
+++ b/app/services/step_factory.rb
@@ -1,80 +1,4 @@
 class StepFactory
-  # need to register permitted names/classes
-  STEP_NAMES = [
-    "Identity",
-    "ReturningTeacher",
-    "PreviousId",
-    "HasTeacherId",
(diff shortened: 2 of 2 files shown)


app/controllers/sitemaps_controller.rb
```diff
@@ -1,7 +1,7 @@
 class SitemapsController < ApplicationController
   def index
     @host = "#{request.protocol}#{request.host}"
-    @steps = StepFactory::STEP_NAMES
+    @steps = Base.descendants.map(&:to_s)
```
jebw, line 4: This would include all child classes, irrespective of whether they are abstract? Whats the advantage of removing the list of what should be there.

Separately - do we actually want to include all the steps in the sitemap?

app/services/step_factory.rb
```diff
@@ -90,10 +14,11 @@ def message
   class << self
     def create(name)
       classified_name = name.camelize
-      # need to redirect to root or last valid step.
-      raise NameNotFoundError, classified_name unless STEP_NAMES.include?(classified_name)
 
       Object.const_get(classified_name).new
```
jebw, line 18: This will now instantiate _any_ ruby class in the system its told to - feels a bit open ended?

jebw (changes requested):
There are downsides to these changes - can you add a PR description explaining why this is being changed

Thanks

ethax-ross:
Good points - it was essentially an effort to reduce the amount of manual work needed to add a new step; as it stands if you add a step you need to know to register it with the factory and then also potentially update the attributes for submission here, though I think this pattern needs to change:

https://github.com/DFE-Digital/get-teacher-training-adviser-service/pull/130/files#diff-537bb3becf881573d17b6c3665496442R1

I'll close this off for now as I'm exploring a refactor that will potentially make this redundant anyway.

PR #157, opened 2020-08-20
Namespace all cache keys in Redis

jebw:
### Context

We wish to share a single Redis instance between both apps, this means we need a way to isolate the keys

1. Namespace all keys under the `TTA` namespace


sfawcett123 (approved):
That looks good

PR #162, opened 2020-08-22
Remove sign up steps from sitemap

ethax-ross:
### Context

It doesn't make sense to expose the sign up steps to be indexed by search engines as they can't be accessed directly; only by progressing through the sign up flow.

Remove sign up steps from sitemap


goodviber (changes requested):
Doesn't this change remove all the sitemap urls? Previously they were generated from the step names

ethax-ross:
Yeah that's exactly what it does - we don't want search engines to index the individual steps as users can't access them directly; they will always be redirected to the start of the journey.

goodviber (approved):

PR #163, opened 2020-08-22
Bug fixes and tidy up for the new TTA journey

ethax-ross:
Fixes a couple of edge case bugs in the new TTA journey, as well as making the steps more readable in general.

### DIFF
--- a/app/models/teacher_training_adviser/steps/gcse_maths_english.rb
+++ b/app/models/teacher_training_adviser/steps/gcse_maths_english.rb
@@ -7,11 +7,12 @@ module TeacherTrainingAdviser::Steps
     OPTIONS = Crm::OPTIONS
 
     def skipped?
-      @store["returning_to_teaching"] ||
-        [
-          TeacherTrainingAdviser::Steps::HaveADegree::DEGREE_OPTIONS[:studying],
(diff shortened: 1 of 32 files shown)


jebw (changes requested):
Can this commit be its own PR - no need for it to be in the big refactor PR

jebw (changes requested):
General purpose fix - can this be in its own PR

goodviber (changes requested):
It looks really good, could we define the methods we are callling within skipped? elsewhere so we would only call then once within the skipped method?

ethax-ross:
I agree we need to share the `skipped?` logic between steps so the complexity isn't compounding as you go further down the journey, but I don't think extracting it from the steps is the right approach here - instead I would put something in the wizard that lets a step query if a previous step was skipped. We could then have 2x QualificationRequired steps (one for maths/english and one for science) where the methods become:

```
def skipped?
  @wizard.skips?(RetakeGcseMathsEnglish)
end

...

def skipped?
  @wizard.skips?(RetakeGcsScience)
end
```

Either way I think this should be picked up as a separate PR once we have it working using the current pattern ^ there's a ticket for it in Jira.

goodviber (approved):

PR #164, opened 2020-08-24
Session expire error

goodviber:
### JIRA ticket number
https://dfedigital.atlassian.net/browse/GITPB-418
### Context
On session expiry we need a graceful failure for ActionController::InvalidAuthenticityToken
Redirect to session-expiry page with post requests with invalid csfr token

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,6 +2,8 @@ class ApplicationController < ActionController::Base
   default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
   before_action :http_basic_authenticate
 
+  rescue_from ActionController::InvalidAuthenticityToken, with: :handle_session_expired
+
 private
 
   def http_basic_authenticate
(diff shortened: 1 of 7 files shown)


app/controllers/application_controller.rb
```diff
@@ -12,4 +14,8 @@ def http_basic_authenticate
         password == ENV["HTTPAUTH_PASSWORD"].to_s
     end
   end
+
+  def handle_session_expired
+    redirect_to root_path
```
ethax-ross, line 19: Can be one for later, but I think it would be worth monitoring how often this happens, either with a log statement or by exposing a metric to prometheus

ethax-ross (approved):

ethax-ross:
I've approved this, but thinking about it - would we want to show a message to the user to let them know what happened?

app/controllers/application_controller.rb
goodviber: have updated with error logging and redirects to session-expiry page, with a link to restart to home page

ethax-ross (approved):

PR #167, opened 2020-08-24
Changed default RSpec logger back to default

jebw:
### Context

The default RSpec logger is more concise when running specs locally. We still need the documentation and SonarQube loggers when running in CI though

1. Drop the hard coded use of the documentation logger
2. Dynamically enable the use of the documentation and sonarqube loggers in CI


goodviber (approved):
lgtm

PR #168, opened 2020-08-25
Issue verification code

ethax-ross:
### JIRA ticket number

[GITPB-415](https://dfedigital.atlassian.net/jira/software/projects/GITPB/boards/61?selectedIssue=GITPB-415)

### Context

When a candidate enters their first/last/email we want to check if we have a matching candidate in the system already. If we do, a verification code should be emailed to the candidate that they can then use to 'authenticate' and we can pre-fill the sign up form.

- Send verification code on matchback

If a candidate enters a first/last/email that matches an existing candidate in the CRM we want to email them a verification code that can be used in a subsequent step to retrieve their details and pre-fill the form.

- Add authenticate step

Add the authenticate step that is shown if a match-back is successful. Allows the user to resend the verification code in case they didn't receive it. The store is pre-filled with the existing candidate information off the back of exchanging the verification code for candidate details.

### Guidance to review

Pretty much all of this is lifted/shifted from the get-into-teaching-app.

Screenshot 2020-08-24 at 14 09 58
Screenshot 2020-08-24 at 14 10 24

### DIFF
--- a/app/controllers/concerns/wizard_steps.rb
+++ b/app/controllers/concerns/wizard_steps.rb
@@ -17,7 +17,7 @@ module WizardSteps
   def update
     @current_step.assign_attributes step_params
 
-    if @current_step.save
+    if @current_step.save!
       redirect_to next_step_path
 
       # Needs to occur after redirect because it purges data after submission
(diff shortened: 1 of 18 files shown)


goodviber (approved):
lgtm

PR #170, opened 2020-08-25
Submit the Candidate information to the API

ethax-ross:
### JIRA ticket number

[GITPB-592](https://dfedigital.atlassian.net/jira/software/projects/GITPB/boards/61?issueParent=19117&selectedIssue=GITPB-592)

### Context

When the candidate has finished filling in the sign up form they hit 'complete', at which point a request should send all the gathered information to the API, which then puts it on a queue to go to the CRM.

- Submit payload to API at end of sign up

When the user completes the sign up we bundle the information they entered into a request that sends it to the API, where it gets queued before going on to the CRM.

### Guidance to review

Mostly a lift/shift from the GiT website - follows the same pattern.


goodviber (approved):
lgtm

PR #173, opened 2020-08-26
Prevent overriding session store key in test environment

ethax-ross:
In the test environment if the session store is overridden it causes the Capybara tests to misbehave; essentially clearing the session between each request (resulting in the sign up flows not working as the order depends on session data from previous steps).

Moves the session store config into development/production environments only.


jebw (approved):
LGTM

PR #174, opened 2020-08-26
Add core wizard classes from GiT app

ethax-ross:
We are going to transition the TTA sign up to use the same format/classes as the GiT app mailing list/event wizards. This commit adds the base/core classes for the Wizard, Store and Step models.

We should look to move these classes into a gem going forward so that we can reuse them between both apps.


jebw (approved):
LGTM

PR #175, opened 2020-08-26
Decouple classes from the ApiClient

ethax-ross:
The ApiClient will be removed along with the existing journey (once the new wizard is in place); instead we will call the GetIntoTeachingApiClient directly (adding the ApiClient is a needless abstraction; originally we thought it may do more, such as caching but that will be done in the gem itself going forward).

It is being decoupled in this commit from any classes/modules that will remain in the app once the old registrations flow is removed.


jebw (approved):
LGTM

PR #177, opened 2020-08-26
Add new TTA journey using the wizard

ethax-ross:
### Context

The current registration flow contains a lot of duplicate steps where the only difference is the contents of the `next_step` method. The models explicitly contain the branching information for every permutation, which makes the model hierarchy verbose: 

`models/start_teacher_training`
`models/equivalent/start_teacher_training`
`models/degree/start_teacher_training`
...etc...

Instead, the steps can be thought of as linear and a branch is simply skipping steps that are not in that particular journey. This is how the GiT app wizards work for mailing list/event sign ups and it results in more concise modelling.

Transitioning the TTA app to use the same wizard logic will have lots of benefits:

- The two apps will be more similar; if you understand how the wizard works you can rationalise both of the apps more easily. We could also extract the shared core/base wizard classes/modules into a shared gem.
- There will only need to be one model per step, instead of one model per step, per branch. This should reduce the complexity and amount of code.
- The answered steps can be determined by iterating over the list of steps, making the review answers step simpler.

- Create models/views/wizard representing linear flow

Re-purposes the existing TTA sign up models and views; consolidating them so the same steps are reused throughout (instead of having a different model for the various branches, as the steps don't actually change). The journey is then controlled by a single wizard, with the steps determining if they should be shown or not depending on the data in the store from previous steps.

- Add controller/views for the wizard

Introduce the `StepsController` to control the Teacher Training Adviser wizard, along with the required views for `show`/`update`/`completed`.

Add feature tests for the core journeys and request specs to ensure it has the `WizardSteps` concern.

- Add custom telephone validator

Currently the telephone validation is done in each model, instead this creates a custom telephone validator and ensures it replicates the API validation logic (allow any non-letter character, between 5 and 20 characters in length).

- Update specs to hit new journey 

### Guidance to review

Best reviewed by-commit; it's a lot of changes but the commits are in logical order.

Although there are a lot of changes here, most of the code is copy/pasted from the existing registrations process (the models were already close to what we need for the linear flow) so the change shouldn't be as big/risky as it perhaps seems on the surface.

I've added a decent amount of feature tests to cover the different TTA journeys and they line up with the current Miro board flow diagram.

I'm not totally happy with the `ReviewAnswers` step at the moment; I've kept it close to how it was in the existing registrations flow to minimise the amount of changes here, but I expect to do a follow up PR to try and make it more generic - so the steps are determining which information appears in the completion flow via a single partial, instead of having conditional logic in lots of different partials (it's going to be slightly awkward as a step can result in multiple rows of information, not always adjacent to each other).

### DIFF
--- a/app/views/teacher_training_adviser/steps/_previous_teacher_id.html.erb
+++ b/app/views/teacher_training_adviser/steps/_previous_teacher_id.html.erb
@@ -1,0 +1,4 @@
+<%= f.govuk_fieldset legend: { text: 'What is your previous teacher reference number?' } do %>
+  <%= f.govuk_text_field :teacher_id, width: 20,
+  label: { text: 'Teacher reference number (optional).' } %>
+  <% end %>
--- a/app/controllers/teacher_training_adviser/steps_controller.rb
+++ b/app/controllers/teacher_training_adviser/steps_controller.rb
@@ -1,0 +1,21 @@
+module TeacherTrainingAdviser
+  class StepsController < ApplicationController
+    include WizardSteps
+    self.wizard_class = TeacherTrainingAdviser::Wizard
+
+  private
+
+    def step_path(step = params[:id])
+      teacher_training_adviser_step_path step
+    end
+    helper_method :step_path
+
+    def wizard_store
+      ::Wizard::Store.new session_store
+    end
+
+    def session_store
+      session[:sign_up] ||= {}
+    end
+  end
+end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -15,4 +15,23 @@ module ApplicationHelper
 
     tag.body attributes, &block
   end
+
+  def govuk_form_for(*args, **options, &block)
+    merged = options.dup
+    merged[:builder] = GOVUKDesignSystemFormBuilder::FormBuilder
+    merged[:html] ||= {}
+    merged[:html][:novalidate] = true
+
+    form_for(*args, **merged, &block)
+  end
+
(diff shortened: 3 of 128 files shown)


app/views/teacher_training_adviser/steps/_previous_teacher_id.html.erb
```diff
@@ -0,0 +1,4 @@
+<%= f.govuk_fieldset legend: { text: 'What is your previous teacher reference number?' } do %>
+  <%= f.govuk_text_field :teacher_id, width: 20,
+  label: { text: 'Teacher reference number (optional).' } %>
+  <% end %>
```
jebw, line 4: Confusing indentation

jebw (approved):
LGTM

There are several tweaks which I think would benefit this but are better as subsequent PRs after this is merged

1. Steps should not duplicate deciding whether an earlier step is skipped - they should ask that earlier step if it was skipped (less sure on this one)
2. The Date parsing needs to go - that should use the ActiveRecord parsing module
3. The partials for the "Check your answers screen" all have a return line in, this could be cleaned up by using a helper method to wrap the render, it can check if the attribute is set, and if it is, call the render of the partial
4. The some of the check your answers partials are checking whether they are in the returning teachers path - the logic of what is included in the check your answers screens should probably be in the Step models and not in the view partials.

ethax-ross:
Agreed 👍  I've started some of those refactoring already in this draft PR https://github.com/DFE-Digital/get-teacher-training-adviser-service/pull/169

PR #178, opened 2020-08-26
Remove legacy TTA journey and direct new sign ups to the new journey

ethax-ross:
### Context

There is a new TTA journey in the app that uses the same wizard infrastructure as the GiT website app; this removes the legacy registration journey and points new sign ups to go to the new journey.

- Point home page to new wizard

Ensures new sign ups go through the new wizard and not the existing/old RegistrationsController.

- Remove the old registration flow

Deletes all classes and modules associated with the current/old registration flow that will be replaced by the new teacher_training_adviser/steps.


jebw (approved):
LGTM

jebw (approved):
Go for it!

PR #179, opened 2020-08-27
Point local and hosted dev environments to new journey

ethax-ross:
We want to be able to access the new journey locally and on the hosted development instance so that it's easy to access and test.


DrLazer (approved):
looks good

PR #182, opened 2020-08-27
Remove third

goodviber:
### Context
We need to remove the third class and pass unknown options from the degree class options
Add helper method to format dropdown with limited options


app/helpers/what_degree_class_helper.rb
```diff
@@ -0,0 +1,5 @@
+module WhatDegreeClassHelper
+  def remove_third_and_grade_unknown(degree_class_options)
+    degree_class_options.reject! { |k| ["Third class or below", "Pass (grade unknown)"].include? k.value }
```
ethax-ross, line 3: We should do this in the model so the validation is correct as well, and we would be better to use the id values in case the copy changes in the CRM.

Since there's only a couple it may be worth doing the same pattern as elsewhere and having an `OPTIONS` constant that explicitly defines the ones we want to include?

goodviber: Sure, have done the OPTIONS hash and validations, but we don't need to move the helper method into the model.

ethax-ross: But can you not then just pass your `OPTIONS` constant into `govuk_collection_select`, making the helper redundant?

goodviber: It's 6 and two threes ...
1. bypass the api call and just hard code the dropdown values, which we could do for the other dropdown models/views
2. we validate against the api call, and the helper trims the available options on display
3. we have a mixture, with the validations hard coded.

If we want to factor in possible future changes to the api, then option 2 will make the code resilient, with the caveat a user could manipulate the dropdown input.

ethax-ross (approved):

PR #184, opened 2020-08-28
Fix review answer display issues

ethax-ross:
A couple of the review answer values either weren't populating due to a type mismatch when keying a hash and another was showing an incorrect value for the `HaveADegree` answer (it was using the value we store, capitalized instead of what we display to the user).


DrLazer (approved):
Looks good

PR #191, opened 2020-09-01
Increased Status Cake Check Frequency

sfawcett123:
Increased Check period to 1minute
Added a workflow to github to deploy


jebw (approved):
LGTM

PR #192, opened 2020-09-01
Include matchback fields in export

ethax-ross:
### JIRA ticket number

[GITPB-684](https://dfedigital.atlassian.net/jira/software/projects/GITPB/boards/61?selectedIssue=GITPB-684)

### Context

When an existing candidate successfully matches back to a contact in the CRM we populate the store with their existing details, including `candidate_id` and `qualification_id`. As we scope the information sent back to the API down to only the attributes of steps that weren't skipped - and `candidate_id`/`qualification_id` are not attributes of any step - the ids
weren't being sent back to the API.

This adds `candidate_id` and `qualification_id` to the `export` call on the `Authenticate` step and also ensures no answers to the `Authenticate` step will ever appear in the review answers view.

- Include matchback fields in export

### DIFF
--- a/app/models/wizard/steps/authenticate.rb
+++ b/app/models/wizard/steps/authenticate.rb
@@ -2,6 +2,7 @@ module Wizard
   module Steps
     class Authenticate < ::Wizard::Step
       IDENTITY_ATTRS = %i[email first_name last_name date_of_birth].freeze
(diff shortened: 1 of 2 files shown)


jebw (approved):
LGTM

PR #193, opened 2020-09-01
Timezone

goodviber:
### JIRA ticket number
https://dfedigital.atlassian.net/browse/GITPB-642
### Context
There are currently a few things to work out with callbacks and timezones:

The time zone isn't sent to the API - do we definitely want to collect it? (we infer if the telephone is international if it starts with a '+' and not '+44'

If we keep timezones, we should change the date/time picker to be in the selected timezone when we show it to the user in both the drop down and on the review answers page

Outcome of meeting with Lizzie:

For overseas we ask them for their telephone and timezone on the current step, then move the callback time slots onto a new step and offset them with the selected timezone from the previous step.
Creation of OverseasTimezone Step, with attributes time_zone and telephone number
### Guidance to review

The review answers page displays the time as UTC, this will be changed to reflect the timezone time, in a separate PR


app/models/teacher_training_adviser/steps/overseas_timezone.rb
```diff
@@ -0,0 +1,24 @@
...
+  class OverseasTimezone < Wizard::Step
+    attribute :time_zone, :string
+    attribute :telephone, :string
+
+    validates :time_zone, presence: { message: "Select a time zone" }
+    validates :telephone, telephone: true, presence: { message: "Enter a telephone number" }
+
+    def reviewable_answers
+      {
+        "time_zone" => time_zone.to_s,
```
ethax-ross, line 11: These are already strings, so shouldn't need to `to_s` them

app/views/teacher_training_adviser/steps/_overseas_timezone.html.erb
```diff
@@ -0,0 +1,14 @@
...
+
+<p class="govuk-body">You need to speak to us and should book a call in advance. We will contact you once your call is booked.</p>
+<p class="govuk-body">You must have the details of your overseas qualifications when we contact you.</p>
+
+<%= f.govuk_phone_field :telephone, width: 20,
+  label: { text: 'Contact telephone number *' } %>
+
+<div class="govuk-form-group">
+  <%= f.label 'Select your time zone', class: "govuk-label" %>
+  <%= f.time_zone_select :time_zone, ActiveSupport::TimeZone.all.drop(1), {}, { class: "govuk-select" }  %>
```
ethax-ross, line 11: Why do we `drop(1)` from the list of timezones?

ethax-ross (commented):
Looks good, couple very minor comments/questions.

app/views/teacher_training_adviser/steps/_overseas_timezone.html.erb
goodviber: 001 > ActiveSupport::TimeZone.all.first
 => #<ActiveSupport::TimeZone:0x00007fc2dd627cb0 @name="International Date Line West", @utc_offset=nil, @tzinfo=#<TZInfo::DataTimezone: Etc/GMT+12>>

app/models/teacher_training_adviser/steps/overseas_timezone.rb
goodviber: 👍

name:
You managed to @ me somewhere in here but I read through, good job keep it up :)

I'll leave now 👋

app/views/teacher_training_adviser/steps/_overseas_timezone.html.erb
ethax-ross: Ah, I see, in that case is it worth putting a method in the model along the lines of:

```
def time_zones_sans_utc
  ActiveSupport::TimeZone.all.drop(1)
end
```

```
f.object.time_zones_sans_utc
```

Then we can write a test for it? feels like we're relying on `ActiveSupport` ordering them a certain way which may change in a future update.

ethax-ross (approved):

PR #197, opened 2020-09-02
Prevent candidate from signing up multiple times

ethax-ross:
### JIRA ticket number

[GITPB-658](https://dfedigital.atlassian.net/jira/software/projects/GITPB/boards/61?selectedIssue=GITPB-658)

### Context

If a candidate has already signed up to the TTA service and successfully matches back, the store will contain a `true` value for `already_subscribed_to_teacher_training_adviser`. When this is the case, we will show them a new step that informs them they have already signed up and directs them to contact support if they think this incorrect.

- Prevent candidate from signing up multiple times

### Guidance to review

Screenshot 2020-09-02 at 14 45 59


jebw (approved):
LGTM

PR #199, opened 2020-09-03
Correctly display callback booking quotas in Candidate's locale

ethax-ross:
### JIRA ticket number

[GITPB-642](https://dfedigital.atlassian.net/jira/software/projects/GITPB/boards/61?selectedIssue=GITPB-642)

### Context

If a Candidate is in the UK then the callback booking quota time slots should be in Europe/London timezone (so GMT or BST). If overseas, then the slots should be converted to the time zone selected by the Candidate. Similarly, on the review answers page the callback date/time needs to align with what was selected (so also in the Candidate’s locale).

Issues with the current conversions:

* Review answers page doesn’t covert to the Candidate’s locale at all currently.
* Select box on callback step currently formats quotas to correct timezone but does not exclude based on the Candidate’s current day (instead it excludes based on UTC day being today).
* Select box on callback step groups quotas by day in UTC instead of converting to the Candidate’s local time zone.

- Rename Timezone to TimeZone

- Convert callback quota slots into Candidate's locale

If a Candidate is overseas we ask them to select their timezone; previously we converted the times but if these fell onto a different day they would still be grouped by the incorrect (UTC) day and an overseas Candidate could end up seeing bookings for their current day (we want to show bookings from tomorrow onwards).

This commit moves more of the booking quota logic into the view helper so we can omit dates on the Candidate's (local) day and correctly group by the day in the Candidate's locale as well.

- Correctly format callback answer in Candidate's time zone


app/helpers/answer_helper.rb
```diff
@@ -6,7 +6,8 @@ def format_answer(answer)
     when Date
       answer = answer.to_formatted_s(:govuk_date)
     when Time
-      answer = answer.to_formatted_s(:govuk_time)
+      time_zone = @wizard.export_data["time_zone"] || "London" # rubocop:disable Rails/HelperInstanceVariable
```
jebw, line 9: `@wizard.export_data` is potentially expensive - its probably better to expose a `timezone` method on the wizard - which in turn uses fetch to retrieve a key from the store.

Its probably not ideal calling a controller variable it from the view helper - in theory they should be unaware of the context outside of their parameters, to allow easy reuse.

I18n.timezone is thread local - it would be better to set it there in a `before_action` in the controller (and back to its original value in an `after_action`

That way all the time zone aware helpers would just 'behave' ?

app/helpers/callback_helper.rb
```diff
@@ -1,11 +1,19 @@
 module CallbackHelper
-  def format_booking_quotas(quotas_by_day, time_zone = "GMT")
-    quotas_by_day.transform_values do |quotas|
-      quotas.map do |quota|
+  def format_booking_quotas(quotas, time_zone = "GMT")
+    quotas_excluding_today = quotas.reject do |quota|
```
jebw, line 3: Can these two blocks get moved to separate helper methods? They're potentially useful elsewhere and it allows testing the individual bits of functionality.

app/views/teacher_training_adviser/steps/_overseas_callback.html.erb
```diff
@@ -2,7 +2,7 @@
 
 <div class="govuk-form-group">
   <%= f.label :phone_call_scheduled_at, "Select your preferred day and time for a callback", class: "govuk-label" %>
-  <%= f.select :phone_call_scheduled_at, format_booking_quotas(TeacherTrainingAdviser::Steps::OverseasCallback::grouped_quotas, @wizard.export_data["time_zone"]), {}, { class: "govuk-select" } %>
+  <%= f.select :phone_call_scheduled_at, format_booking_quotas(GetIntoTeachingApiClient::CallbackBookingQuotasApi.new.get_callback_booking_quotas, @wizard.export_data["time_zone"]), {}, { class: "govuk-select" } %>
```
jebw, line 5: Whats the logic for moving data layer work into the View? I'd prefer to see this remain in the data model. This means it can be tested more easily, and if additional manipulation on the data is required in a future iteration, it can then be done in the model?

jebw (changes requested):
I'd prefer to see the API work happen in the data models (possibly with a module if it needs sharing between Step models). That would also avoid needing to data filtering in the helpers?

With regard to the timezone conversion - I think thats better handled by changing the global (thread level) timezones - then everything is expecting to parse and export in that timezone unless told otherwise?

app/views/teacher_training_adviser/steps/_overseas_callback.html.erb
ethax-ross: Yeah it should be - it's a bit mixed in the app as a whole right now; I created a ticket last week to go through it all and make it consistent https://dfedigital.atlassian.net/browse/GITPB-646

app/helpers/answer_helper.rb
ethax-ross: That's a nice suggestion, I'll look into configuring the timezone for the context of the request 👍

app/models/concerns/callback_booking_quotas.rb
```diff
@@ -0,0 +1,8 @@
+module CallbackBookingQuotas
+  def callback_booking_quotas
+    quotas = GetIntoTeachingApiClient::CallbackBookingQuotasApi.new.get_callback_booking_quotas
+    quotas.reject do |quota|
+      quota.start_at.in_time_zone.to_date == Time.zone.now.to_date
```
jebw, line 5: minor - but this could be `.today?`

app/views/teacher_training_adviser/steps/_overseas_callback.html.erb
```diff
@@ -2,7 +2,7 @@
 
 <div class="govuk-form-group">
   <%= f.label :phone_call_scheduled_at, "Select your preferred day and time for a callback", class: "govuk-label" %>
-  <%= f.select :phone_call_scheduled_at, format_booking_quotas(TeacherTrainingAdviser::Steps::OverseasCallback::grouped_quotas, @wizard.export_data["time_zone"]), {}, { class: "govuk-select" } %>
+  <%= f.select :phone_call_scheduled_at, callback_options(f.object.class.callback_booking_quotas), {}, { class: "govuk-select" } %>
```
jebw, line 5: :ok_hand:

jebw (changes requested):
Looks really good, feels much cleaner now

Suggested tweaks
1. Use rails own time helpers instead of time cop to avoid a dependency
2. Use `.today?` to do the comparison for today, not at all fussed by this if you think its more obvious as it is
3. Clean up the Time formatting in answer_helper - I'm really not keen on having the helper know about the internal data structures of the wizard's store.

Gemfile.lock
```diff
@@ -294,6 +294,7 @@ GEM
       sprockets (>= 3.0.0)
     thor (1.0.1)
     thread_safe (0.3.6)
+    timecop (0.9.1)
```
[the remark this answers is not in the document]
ethax-ross: Nice, didn't realise that 👍

jebw (approved):
LGTM

PR #200, opened 2020-09-03
Display review answers correctly when steps have been missed

ethax-ross:
### JIRA ticket number

[GITPB-688](https://dfedigital.atlassian.net/jira/software/projects/GITPB/boards/61?selectedIssue=GITPB-688)

### Context

It is currently possible for a user to miss a step; you can jump straight to the review answers step if you have the URL (though if you try and complete the submission it will redirect to the first invalid step).

If a candidate does this it causes an exception due to not expecting a non-skipped step to have a `nil` value (which happens if they manually plug a future step URL in). This commit updates the `reviewable_answers` to ensure an exception will never be raised, instead displaying an empty answer in the review answers step for any they have missed.

A better long-term solution may be to redirect to an invalid step if the candidate attempts to access a future step manually via the URL.

- Display review answers correctly when steps have been missed

### DIFF
--- a/app/models/teacher_training_adviser/steps/have_a_degree.rb
+++ b/app/models/teacher_training_adviser/steps/have_a_degree.rb
@@ -24,7 +24,7 @@ module TeacherTrainingAdviser::Steps
 
     def reviewable_answers
       {
-        "degree_options" => I18n.t("have_a_degree.degree_options.#{degree_options}"),
+        "degree_options" => degree_options ? I18n.t("have_a_degree.degree_options.#{degree_options}") : nil,
       }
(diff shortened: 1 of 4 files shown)


jebw (approved):
I think the user _should_ be able to skip to the review answers screen provided then can not submit

The more smarts around state that go in, the harder it will be to maintain. I think so long as its being check for correctness prior to submission, then that is sufficient?

ethax-ross:
Yeah true, it should be an edge-case either way really. I wonder if we should be no-indexing the wizard steps though; maybe a question for PM.

PR #206, opened 2020-09-07
Feature/validation message changes

DrLazer:
"Select an option"
is not enough context in a validation message for screen reader users

Updated all the validation messages where this phrase is used and had Darren look over the wording of them

https://trello.com/c/h3qvtwyl/292-tta-flow-you-must-select-either-yes-or-no

### DIFF
--- a/app/models/teacher_training_adviser/steps/retake_gcse_maths_english.rb
+++ b/app/models/teacher_training_adviser/steps/retake_gcse_maths_english.rb
@@ -2,7 +2,7 @@ module TeacherTrainingAdviser::Steps
   class RetakeGcseMathsEnglish < Wizard::Step
     attribute :planning_to_retake_gcse_maths_and_english_id, :integer
 
-    validates :planning_to_retake_gcse_maths_and_english_id, types: { method: :get_candidate_retake_gcse_status, message: "Select yes if you are planning to retake English or Maths (or both) GCSE or equivalent" }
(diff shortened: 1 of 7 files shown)


app/models/teacher_training_adviser/steps/retake_gcse_maths_english.rb
```diff
@@ -2,7 +2,7 @@ module TeacherTrainingAdviser::Steps
   class RetakeGcseMathsEnglish < Wizard::Step
     attribute :planning_to_retake_gcse_maths_and_english_id, :integer
 
-    validates :planning_to_retake_gcse_maths_and_english_id, types: { method: :get_candidate_retake_gcse_status, message: "You must select either yes or no" }
+    validates :planning_to_retake_gcse_maths_and_english_id, types: { method: :get_candidate_retake_gcse_status, message: "Select yes if you are planning to retake either English or maths (or both) GCSEs, or equivalent" }
```
ethax-ross, line 5: Is the casing not a bit off in some of theses? I would have expected this to be:

```...either English or Maths (or both) GCSEs, or equivalent```

or

```...either english or maths (or both) GCSEs, or equivalent```

DrLazer: copied it from the question, not sure, Darren approved them though

DrLazer: could be wrong in the question also

ethax-ross (approved):

PR #211, opened 2020-09-09
Format privacy policy with allowed HTML tags

ethax-ross:
### JIRA ticket number

[GITPB-590](https://dfedigital.atlassian.net/jira/software/projects/GITPB/boards/61?selectedIssue=GITPB-590)

### Context

The privacy policy in the CRM has been updated to contain HTML for bullet points, anchor links, etc. We want this to be correctly rendered in the app.

- Format privacy policy with allowed HTML tags

### DIFF
--- a/spec/helpers/text_format_helper_spec.rb
+++ b/spec/helpers/text_format_helper_spec.rb
@@ -43,5 +43,10 @@ RSpec.describe TextFormatHelper, type: :helper do
       let(:html) { "<script>malicious</script>" }
       it { is_expected.to eql "malicious" }
     end
+
+    context "with malicious anchor tags" do
+      let(:html) { "<a href=\"http://test.com\" onclick=\"somethingNasty();\">boom</a>" }
+      it { is_expected.to eql "<a href=\"http://test.com\">boom</a>" }
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


spec/helpers/text_format_helper_spec.rb
```diff
@@ -30,4 +30,18 @@
...
+
+  describe "#safe_html_format" do
+    subject { safe_html_format html }
+
+    context "with allowed HTML" do
+      let(:html) { "<p><strong>hello</strong> <a href=\"http://test.com\">world</a></p><ul><li>test</li></ul>" }
+      it { is_expected.to eql html }
+    end
+
+    context "with disallowed HTML" do
```
jebw, line 42: Can we add a test in here for malicicious attributes, eg `<a onclick` - I can see the code handles this but would be good to have a test for that behaviour?

jebw (changes requested):
Could do with a check for malicious `<a` tags

spec/helpers/text_format_helper_spec.rb
ethax-ross: Yeah good call, I'll do that on both PRs

jebw (approved):
LGTM

PR #212, opened 2020-09-09
Feature/page titles

DrLazer:
titles are duplicated throughout the TTA and causing DAC failures

https://trello.com/c/h85wPqf8/280-tta-flow-duplicated-page-titles

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -10,7 +10,7 @@ Rails.application.configure do
   config.eager_load = false
 
   # Show full error reports.
-  config.consider_all_requests_local = true
+  config.consider_all_requests_local = false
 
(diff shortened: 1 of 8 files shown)


config/environments/development.rb
```diff
@@ -10,7 +10,7 @@
   config.eager_load = false
 
   # Show full error reports.
-  config.consider_all_requests_local = true
+  config.consider_all_requests_local = false
```
ethax-ross, line 13: This should be switched back to `true` by default in dev

ethax-ross (commented):
Looks good, just needs tests 😉

ethax-ross (approved):

PR #236, opened 2020-09-18
Remove current year after 17th September

ethax-ross:
Once it has passed the 17th September it is no longer possible to apply for teacher training for the current year.


DrLazer (approved):
looks good

PR #241, opened 2020-09-21
update return to git link

DrLazer:
update return to git link


jebw (changes requested):
This needs to use a per-environment configuration to link back so we get different URLs depending upon the environment

ethax-ross (approved):

jebw (approved):

PR #242, opened 2020-09-21
update link on get the right gcse page

DrLazer:
update link on get the right gcse page


jebw (changes requested):
This needs to use an per environment configuration to link back to the correct version of the main website

jebw (approved):

PR #245, opened 2020-09-22
Remove No Preference from subject taught list

ethax-ross:
We should not allow 'no preference' as an option for previous subject taught.


jebw (approved):
LGTM

PR #267, opened 2020-09-25
Point preprod to test API instance

ethax-ross:
The CRM team want to be able to test against the test CRM - going forward this needs to change to test anyway.


sfawcett123 (approved):
LGTM

PR #270, opened 2020-09-25
Contract tests

mattonfoot:
### Component tests
Component tests are designed to prove that an inbound request to the component will trigger the correct outbound requests from the component to its dependencies. This allows the dependencies to then prove that they handle the outbound requests correctly within their test suite.

This pull request adds a component test spec file to the application test suite which iterates on the generated test fixtures coming from the contract test fixture repository

### Guidance to review
The test should only run if the appropriate files are found. If you wish the tests to run, then the contents of the contracts folder from the test fixture repository should be added to tmp/contracts in this folder structure.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 # Ignore all logfiles and tempfiles.
 /log/*
 /tmp/*
+/contracts/*
 
 # Ignore uploaded files in development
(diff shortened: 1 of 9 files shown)


sfawcett123:
Closing will leave branch

PR #279, opened 2020-09-29
Created internal routes for Prometheus

sfawcett123:
To enable Prometheus to scrape the metrics from each node of the adviser, internal routes are required.

The routes have had to be removed from the application.routes method and moved to routes.application.  Since there were now more resources it made sense to put routes into its own terraform file. (routes.tf)


jebw (approved):
LGTM

PR #282, opened 2020-09-29
Feature/689/logit integration

jebw:
### JIRA ticket number

GITPB-689

### Context

We should be generating logs which are useful to LogIt - initially that means single line logs, or logs where the multiple parts of a request can be grouped togther

1. Make STDOUT logs synchronous
2. Only load Semantic logging if either SEMANTIC_LOGGING or CF_INSTANCE_GUID are set
3. Make puma reopen the log output when if it forks a worker if running with semantic logging

### Notes

1. I've tested to make sure param filtering is honoured - visiting `http://localhost:3000/?first_name=Joe` shows as `[FILTERED]`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -84,5 +84,12 @@ group :test do
   gem "webdrivers", "~> 4.3"
 end
 
+group :rolling, :preprod, :userresearch, :production do
+  # loading the Gem monkey patches rails logger
+  # only load in prod-like environments when we actually need it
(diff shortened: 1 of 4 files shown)


ethax-ross (approved):

peteryates (approved):
Looks good.

sfawcett123:
Closing it has got old and I dont think it will progress now,

PR #284, opened 2020-10-05
Remove No Preference and Other from SubjectInterestedTeaching step

ethax-ross:
We no longer want to allow the options 'No Preference' and 'Other' to be selected in response to the question "What subject would you like to teach?".

### DIFF
--- a/app/views/teacher_training_adviser/steps/_subject_interested_teaching.html.erb
+++ b/app/views/teacher_training_adviser/steps/_subject_interested_teaching.html.erb
@@ -1,6 +1,6 @@
 <%= f.govuk_collection_select :preferred_teaching_subject_id,
-  GetIntoTeachingApiClient::TypesApi.new.get_teaching_subjects,
-  :id,
-  :value,
+  f.object.class.options,
+  :last,
+  :first,
   label: { text: "Which subject would you like to teach?", tag: "h1", size: "m" }
 %>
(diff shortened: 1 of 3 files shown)


app/views/teacher_training_adviser/steps/_subject_interested_teaching.html.erb
```diff
@@ -1,6 +1,6 @@
 <%= f.govuk_collection_select :preferred_teaching_subject_id,
-  GetIntoTeachingApiClient::TypesApi.new.get_teaching_subjects,
-  :id,
-  :value,
+  f.object.class.options,
```
peteryates, line 2: Is there a benefit to getting at the class this way versus `SubjectInterestedTeaching.options`?

peteryates (approved):
It looks fine to me. Left a comment but there's probably a good reason for the approach.

app/views/teacher_training_adviser/steps/_subject_interested_teaching.html.erb
jebw: I'd say yes - this means if the class is renamed, than the form wont require changes

jebw (approved):
LGTM

app/views/teacher_training_adviser/steps/_subject_interested_teaching.html.erb
ethax-ross: The only real benefit is that we won't need to change this line if the model name changes so it should make refactoring simpler. I can see an argument that having the class name here may improve readability, but it's a bit six and two threes in my mind.

peteryates: Yeah - makes sense. I was originally thinking of it from the 'where is this used?' point of view; easier to search for it by class. Happy either way though

PR #289, opened 2020-10-06
Workflow Pass in Tags

sfawcett123:
When releasing to Production we need to be able to re-release older versions. This means we need to pass in a tag.

However, we do not want to release tags that have not been tested so, we have a check that ensures a tag has been part of a release.

I have tested this by:

-  Commenting out all the code that does anything ( terraform, sentry and slack )
-  Running the action from the branch devops/workflows/pass-tag-in with a tag of "master" . As expected Failed
-  Running the action from the branch devops/workflows/pass-tag-in with a tag of "BETA-1.0". As expected Passed
-  Remove comments

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -2,6 +2,10 @@
 name: Release to Production
 on: 
   workflow_dispatch:
+    inputs:
+      tags:
+        description: 'Release Tag'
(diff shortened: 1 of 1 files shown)


jebw (approved):
LGTM

PR #290, opened 2020-10-06
Update CSP for doubleclick errors

jebw:
### Context

We're getting ad.doubleclick.net errors from CSP

Allow ad.doubleclick.net as an image source


ethax-ross (approved):
These third party scripts requiring `unsafe-*` all but make the CSP more hassle than it's worth 😩

peteryates:
Code looks fine, `unsafe-eval` sets some alarm bells off though.

jebw:
Yes completely agree - really don't want to add unsafe-eval, but alternatives are we don't add the analytics or we get spammed from them (and they probably don't work). If we drop the analytics in the future I think we should be starting with a clean slate for the CSP

PR #294, opened 2020-10-08
Add missing validation messages

ethax-ross:
- Add missing validation messages

Some attribute validation rules didn't have corresponding custom error messages; this introduces them to the translations file.

- Remove redundant validation checks

These attributes are not set by the user; instead they are set by the model internally. If these validation checks ever failed we would end up showing an error message to the user that they could do nothing about, and we would not be alerted to a bug in the application code.

Instead, by removing the validation checks it will enable the candidate to get to the end of the submission and they will then receive an error when the form attempts to submit to the API. We will also be alerted about this, which feels like a better outcome/experience for both the candidate and ourselves.


config/locales/en.yml
```diff
@@ -108,10 +108,17 @@ en:
           attributes:
             first_name:
               blank: "You need to enter your first name"
+              too_long: "You have entered too many characters - you need to enter your first name"
             last_name:
               blank: "You need to enter your surname"
+              too_long: "You have entered too few characters - you need to enter your surname"
```
peteryates, line 114: I think this should be `too many`

peteryates (commented):
Looks good but there's a typo in the `last_name.too_long` error message.

config/locales/en.yml
ethax-ross: Yep, well spotted!

peteryates (approved):
:+1:

PR #296, opened 2020-10-08
Minor bug fixes after error message refactoring

ethax-ross:
- Correct typo in error message

- Use entity value for both option value and name

On the `WhatSubjectDegree` step we send the text value of the subject instead of the GUID (which is unique to this step - the field expects a text value instead of ID).


sfawcett123 (approved):
OK

PR #297, opened 2020-10-08
Verify request payload in feature specs

ethax-ross:
A bug was introduced whereby the `WhatSubjectDegree` step was incorrectly persisting the GUID value of the subject instead of the text value.

This PR adds test coverage to the request payload sent in feature specs, to ensure the above scenario - along with other
potential bugs - are caught.


peteryates (approved):
Looks good. I'd never seen the `having_attributes` syntax before, neat.

PR #299, opened 2020-10-09
Clean up HaveADegree step

ethax-ross:
Improve naming of options to better match what is shown to the user.

Remove unused constants.

Remove redundant validation check for date being in the past; this is covered on the specs for the `years` method/inclusion validation.


peteryates (approved):
Looks good to me

PR #300, opened 2020-10-09
Clean up step skipped? logic

ethax-ross:
- Simplify skipped? methods

Simplifies the `skipped?` logic in steps by allowing them to query the skipped state of previous steps.

The tests have only been refactored where necessary.

- Simplify skipped? tests

As the `skipped?` logic is now more concise the tests can also be made simpler.

### DIFF
--- a/app/models/teacher_training_adviser/steps/gcse_maths_english.rb
+++ b/app/models/teacher_training_adviser/steps/gcse_maths_english.rb
@@ -13,7 +13,7 @@ module TeacherTrainingAdviser::Steps
     end
 
     def skipped?
-      @wizard.all_skipped?(WhatSubjectDegree.key)
+      other_step(:what_subject_degree).skipped?
     end
   end
 end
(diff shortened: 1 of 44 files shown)


jebw (commented):
I'm looking over this - in the abstract its a great clean up but I wonder if it could be simplified further.

something like 

```ruby
class MyStep < Step
  def other_step(key_or_class)
    key = key_or_class.kind_of?(Step) ? key_or_class.key : key_or_class
    @wizard.find(key)
  end
end
```

This would simplify accessing other steps, and simplify a lot of the other code


I think its good to reference other steps - I think its fine to use the key directly though rather than chaining through the class to get the key

The `@wizard.all_steps` feels a little indirect, it would read more simply as 

```ruby
def skipped?
  other_step(:uk_postcode).skipped? && other_step(:uk_phonenumber).skipped?
end
```
-----------

Thinking further on this - is it worth having some type of `needs` method which takes keys of 1 or more earlier steps, that would do the same job as `@wizard.all_skipped?` but would be more declarative - thinking like the needs key in the YAML pipelines in GitHub workflows

peteryates (approved):
On the face of it it looks sensible

ethax-ross:
I've pushed another commit to simplify the logic around getting other steps as suggested, and I think there's more that can be done still to simplify it even further but that should be picked up in a follow up PR IMO as this one is big enough.

jebw (approved):
Great clean up!

jebw (approved):

PR #301, opened 2020-10-09
Fix missing translation for HaveADegree step

ethax-ross:
Yaml was translating yes/no keys into `true` and `false`, which mean the translation key `have_a_degree.degree_options.yes` was not matching. By wrapping it in quotes YAML respects the yes/no key values.


peteryates (approved):
Ah, YAML :facepalm:

PR #302, opened 2020-10-10
Remove 'No Preference' from WhatSubjectDegree step

ethax-ross:
We don't want to display the 'No Preference' option for teaching subjects in the `WhatSubjectDegree` step, as it doesn't make sense as an answer.


peteryates (approved):
Looks good

PR #306, opened 2020-10-14
Added google adservice to CSP

jebw:
### JIRA ticket number

N/a

### Context

We are seeing CSP errors from googles add service pixel being blocked

Added an exception for Googles Ad Service


peteryates (approved):
Looks good. Perhaps it's worth creating another array of google services that can be dropped in like with `google_analytics`?

PR #309, opened 2020-10-14
API requires the teaching subject set for primary applicants

jebw:
### JIRA ticket number

N/a

### Context

We should be including the primary teaching subject as the preferred teaching subject id when submitting primary stage applicants to the API

1. This includes the expected primary subject in the export hash from the teaching stage. This only occurs if the teaching stage is skipped.

### Guidance to review

I've since thought of a better way of doing this but this approach is reliable and I can shift things around later on.


peteryates (approved):
Makes sense with regards to the reported bug :+1:

PR #351, opened 2020-10-26
Render 404 on Wizard::UnknownStep

ethax-ross:
We currently fail with a 500 error if the wizard is initialised with an unknown step; this happens if the user enters a bad URL, so it would make more sense to return a 404 in this scenario.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,8 +3,17 @@ class ApplicationController < ActionController::Base
   before_action :http_basic_authenticate
 
   rescue_from ActionController::InvalidAuthenticityToken, with: :session_expired
+  rescue_from ActionController::RoutingError, with: :render_not_found
+
+  def raise_not_found
+    raise ActionController::RoutingError, "Not Found"
+  end
 
 private
+
+  def render_not_found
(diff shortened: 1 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -3,9 +3,18 @@ class ApplicationController < ActionController::Base
   before_action :http_basic_authenticate
 
   rescue_from ActionController::InvalidAuthenticityToken, with: :session_expired
+  rescue_from ActionController::RoutingError, with: :render_not_found
+
+  def raise_not_found
+    raise ActionController::RoutingError, "Not Found"
```
peteryates, line 9: Is it worth adding a bit more context to the raised error, like what's not been found? If it's already visible elsewhere then ignore this

peteryates (approved):
:+1:  minor comment but looks good

app/controllers/application_controller.rb
ethax-ross: I think wherever this is used it should always be inferable/have enough context from the URL, so where I currently have it you would get a 404 on a path like `/teacher_training_adviser/sign_up/some-wrong-step`

PR #354, opened 2020-10-29
Add Credentials to Docker Deploy

sfawcett123:
Github are rate limiting anonymous pulls, this may effect the UK Gov Paas, so by adding credentials that stops.
Also ran **terraform fmt**


peteryates (approved):
lgtm

PR #356, opened 2020-10-30
Rate limit transactions

ethax-ross:
### JIRA ticket number

[Trello-352](https://trello.com/c/MjWfjLi7/352-set-rate-limiting-on-api)

### Context

The API is rate limited to 30 requests/client API key/minute for transactional/vulnerable endpoints (both apps currently share the same API key, but this will change in the future).

We want to apply per-user/IP rate limiting in the front end applications as well, to add more granular control over submissions. The rate limits initially will be 5 requests/IP/minute - this should be plenty for now, but may need tweaking as we roll out to more users (or to take into account advertising 'spikes' in traffic).

When a user encounters a rate limit we want to display a 429 error page similar to the existing 404/500 error pages.

If the user encounters an API rate limit we want to display the same page.

- Rate limit transaction endpoints

The transaction endpoints are rate limited to 30 requests per minute by the API. We want to enforce the rate limiting client-side per user/ip as well to a maximum of 5 requests/IP/minute.

If a user encounters a rate limit, instead of returning a standard 429 and text response we instead render a 'Too many requests' error page, similar to the existing 404 and 500 error pages.

- Rescue ApiError and render 429 error page

Whilst the app itself is rate limited the API also has rate limiting. In the unlikely event that the client rate limits don't trigger and instead the API's rate limiting is enforced we want to display the same 429 error page to the user.

Raise `ApiError` up to the `ApplicationController` when the `code` is 429. The `ApplicationController` will then rescue the error and render the error page.

### Guidance to review

Screenshot 2020-10-30 at 09 03 12

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -19,11 +19,11 @@ private
   end
 
   def render_too_many_requests
-    render template: "errors/too_many_requests", status: :too_many_requests
+    render template: "errors/too_many_requests", status: :too_many_requests, layout: "application"
   end
 
   def render_not_found
-    render template: "errors/not_found", status: :not_found
+    render template: "errors/not_found", status: :not_found, layout: "application"
   end
 
   def http_basic_authenticate
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,12 +4,23 @@ class ApplicationController < ActionController::Base
 
   rescue_from ActionController::InvalidAuthenticityToken, with: :session_expired
   rescue_from ActionController::RoutingError, with: :render_not_found
+  rescue_from GetIntoTeachingApiClient::ApiError, with: :handle_api_error
 
   def raise_not_found
     raise ActionController::RoutingError, "Not Found"
   end
 
 private
+
+  def handle_api_error(error)
+    render_too_many_requests && return if error.code == 429
+
+    raise
+  end
+
+  def render_too_many_requests
+    render template: "errors/too_many_requests", status: :too_many_requests
+  end
 
   def render_not_found
     render template: "errors/not_found", status: :not_found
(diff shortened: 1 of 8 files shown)


app/controllers/application_controller.rb
```diff
@@ -4,13 +4,24 @@ class ApplicationController < ActionController::Base
...
+  rescue_from GetIntoTeachingApiClient::ApiError, with: :handle_api_error
 
   def raise_not_found
     raise ActionController::RoutingError, "Not Found"
   end
 
 private
 
+  def handle_api_error(error)
+    render_too_many_requests && return if error.code == 429
```
jebw, line 16: Feels like it would be nicer to identify this as a 429 within the client, and raise a specific descriptive exception (eg `GetIntoTeachingApiClient::RateLimitExceededError`) in the client. Then you could rescue that particular exception here?

app/controllers/application_controller.rb
```diff
@@ -4,13 +4,24 @@ class ApplicationController < ActionController::Base
...
+
+    raise
+  end
+
+  def render_too_many_requests
+    render template: "errors/too_many_requests", status: :too_many_requests
```
jebw, line 22: This should specify the layout as well - I got bitten by this previously - template inheritance meant it defaulted to the wrong layout for certain pages if the descendant controller has the same name as a layout file.

jebw (changes requested):
This feels like there's two changes here, 

1. A set of rate limiting controls with associate specs
2. Better handling of rate limit exceptions 

it may be simpler to separate them out but not strong preferences.

It would also be nicer to raise less generic errors from the API client? That would benefit both applications?

app/controllers/application_controller.rb
ethax-ross: I agree that would be a little cleaner, but IMO the fact we'd have to add to the monkey patching to achieve it outweighs the benefit it brings. I'd prefer to limit the amount of patching we're doing if possible, but if you feel strongly about it happy to discuss?

ethax-ross:
I've updated the layout to make it explicit. I agree it could have been two PRs, but it's cleanly separated as individual commits so I don't think you gain anything by breaking it up now... the overall PR isn't all that big; a fair amount of it is repetition of rack attack rules.

app/controllers/application_controller.rb
jebw: Ok, lets leave it for now.

jebw (approved):

PR #367, opened 2020-11-04
Prevent error when skipping ahead to review answers step

ethax-ross:
### JIRA ticket number

[Trello-518](https://trello.com/c/oaWXG1mL/518-fix-sentry-issue-in-tta)
[Sentry](https://sentry.io/organizations/dfe-bat/issues/2006318159/?project=5276960&referrer=slack)

### Context

If a user proceeds through the TTA sign up as a candidate with an equivalent degree that is located in the UK, but then manually enters the URL for the review answers page before completing the `UkCallback` step an error would be thrown as the `phone_call_scheduled_at` is `nil` and we try and call `to_date` on it.

- Prevent error when skipping ahead to review answers step

Update to use the safe navigation operator on `phone_call_scheduled_at`.

### DIFF
--- a/app/models/teacher_training_adviser/steps/uk_callback.rb
+++ b/app/models/teacher_training_adviser/steps/uk_callback.rb
@@ -14,8 +14,8 @@ module TeacherTrainingAdviser::Steps
 
     def reviewable_answers
       {
-        "callback_date" => phone_call_scheduled_at.to_date,
-        "callback_time" => phone_call_scheduled_at.to_time, # rubocop:disable Rails/Date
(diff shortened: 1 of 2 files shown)


jebw (approved):
Approving as an immediate fix

ethax-ross:
I did a quick scan over the other steps and it's only this one that has the potential to raise an error; the other steps will evaluate to `nil` if they haven't yet been answered. I could write a test that mocks out `skipped?` to be `false` for all steps and loads the `review_answers` step though, that would cover them all -- I'll raise a PR for that now.

PR #368, opened 2020-11-04
Add test coverage for reviewing incomplete steps

ethax-ross:
A user is able to skip straight to the review answers page manually by entering the correct URL without actually completing any steps. This is expected behaviour, however we need to ensure the steps can handle this scenario and don't assume they have been answered when their `reviewable_answers` method is called.

Add test to simulate rendering all wizard steps for review as unanswered.


jebw (approved):
LGTM

PR #375, opened 2020-11-09
Added a save button to the Cookie Preferences page

jebw:
### Context

The cookie preferences should require the user to save their preferences rather than auto-saving changes

1. Add a save buttom
2. Added a back link which will trigger a page reload
3. Replaced existing back link with one which will reload the page

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -60,4 +60,12 @@ module ApplicationHelper
   def link_to_git_site(text = "Get into Teaching", attributes = {})
     link_to text, (ENV["GIT_URL"].presence || "/url-not-set"), attributes
   end
+
+  def internal_referer
+    referer = request.referer
+    internal = referer.to_s.include?(root_url)
+    return nil unless internal
+
+    referer
+  end
 end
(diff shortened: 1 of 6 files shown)


app/helpers/application_helper.rb
```diff
@@ -60,4 +60,12 @@ def link_to_change_answer(step)
   def link_to_git_site(text = "Get into Teaching", attributes = {})
     link_to text, (ENV["GIT_URL"].presence || "/url-not-set"), attributes
   end
+
+  def internal_referer
+    referer = request.referer
+    internal = referer.to_s.include?(root_url)
+    return nil unless internal
```
peteryates, line 67: is this short enough to just be `return nil unless referer.to_s.include?(root_url)`

peteryates (approved):
Looks good overall. One small comment but no strong feelings on it.

app/helpers/application_helper.rb
jebw: Yeh, agreed - its a cut and paste from `-app` so I'll leave it as is for cherry-picking for now - plan is to move it all to common code anyway

PR #379, opened 2020-11-11
Simplify PolicyValidator

ethax-ross:
We don't need to pass the `method:` into this validator as it always calls `get_privavcy_policy`.

Only rescue on 400/bad request; this indicates the policy entered does not exist in the system. Other errors will bubble up to show the respective error page to the user.


jebw (approved):
LGTM

sfawcett123 (commented):
could 0a203956-e935-ea11-a813-000d3a44a8e9 change?

ethax-ross:
The ID won't change as it's recorded in one of the VCR casettes

PR #386, opened 2020-11-17
Renamed Workflow

sfawcett123:
Trial workflow needs to be correct name in master or it cannot be picked up by the preceding workflow, even though it isn't running on the master branch

### Trello card
[Continuous Deployment](https://trello.com/c/UQrm3g59/516-continuous-deployment)

### Context
Workflow needs to exist in master for the system to even consider working on branches

### Guidance to review
This is a single file that should not effect anything, it isn't fully working but because of how github works, it is a place holder.

### DIFF
--- a/.github/workflows/new_deploy.yml
+++ b/.github/workflows/new_deploy.yml
@@ -1,4 +1,4 @@
-name: Deploy
+name: Deploy to PaaS
 on:
   workflow_dispatch:
     inputs:


jebw (changes requested):
The workflow filename implies this is about creating a new server setup? Is that right?

If so should the name be something more explicit like "New deploy to PaaS" ?

sfawcett123:
> The workflow filename implies this is about creating a new server setup? Is that right?
> 
> If so should the name be something more explicit like "New deploy to PaaS" ?

No, it is the new deployment script, it will be renamed back to deploy.yml in the end, but I want to keep it separate from the running script, because I have to have the workflow in master and cannot name it the same as the one it ill eventually replace

jebw (approved):

PR #387, opened 2020-11-17
Devops/workflows/automated/continuous deployment

sfawcett123:
### Trello card
[Continuous Deployment](https://trello.com/c/UQrm3g59/516-continuous-deployment)
### Context
Enable the automatic deploy of jobs to QA and then through to PRODUCTION

This is an initial stage, we are breaking down the pipeline into phases that can be parameterised and reused.  So the **Deploy** phase will eventually run the same code regardless of environment and will also allow for application only deploys to a specific space. which enables us to deploy review applications

### Guidance to review
This is a complex review and could break the delivery pipeline if there is an unforseen issue, specially since running the code external to github is virtually impossible and running on a branch isnt the same as running in master.


jebw (approved):
LGTM

PR #390, opened 2020-11-18
Deploy for Review

sfawcett123:
### Trello card
[Platform: Continuous Deployment](https://trello.com/c/UQrm3g59/516-platform-continuous-deployment)
### Context
Add the steps necessary to review PRs on a Review server. PO than confirms merge request and that triggers a deploy through the DEV, TEST and PROD environments.

Build Pipeline changed so that when a PR is opened a terraform deploy is executed which will deploy the TTA to a specific name, and post that name in SLACK and the PR.
When the PR is closed the review application is destroyed. 

### Guidance to review
This isn't a totally finished job and needs to be improved, but the basics need to be deployed, so we can see how it works in reality


ethax-ross (approved):
Struggling to follow but _seems_ fine from what I can tell 👍

PR #391, opened 2020-11-18
Improve user experience when SubjectLikeToTeach is not found

ethax-ross:
### Trello card

[Trello-564](https://trello.com/c/w3emKbaL/564-on-returners-step-5-of-tta-service-amend-cant-see-my-subject-and-include-better-cta)

### Context

We want to improve the visibility of the "Can't see the subject you would like to teach?" prompt by changing it to a radio button. If selected, we should take the candidate to a new page with improved copy and calls to action.

Add GiT link helpers for events/mailing list

We need to be able to link to the mailing list and events as CTAs, this adds convenience helpers.

- Improve user experience when SubjectLikeToTeach is not found

Change the "Can't see the subject you would like to teach?" prompt to be an extra radio button on the "What subject would you like to teach if you return to teaching?" question.

Add an additional step to handle when the user selects the new radio option (similar to the `QualificationRequired` step).

Tidy up partial for rendering radio buttons.

Improve wording of error message to make sense with the new option value.

Extend the `ApiOptions` helper with support for including only a subset of the types returned from the API.

### Guidance to review

Screenshot 2020-11-19 at 08 12 26
Screenshot 2020-11-23 at 10 17 44
Screenshot 2020-11-19 at 08 11 42

### DIFF
--- a/app/views/teacher_training_adviser/steps/_subject_not_found.html.erb
+++ b/app/views/teacher_training_adviser/steps/_subject_not_found.html.erb
@@ -1,0 +1,13 @@
+<main class="govuk-main-wrapper no-margin" id="main-content" role="main">
+  <div class="govuk-width-container">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        <h1 class="govuk-heading-l">Get support</h1>
+
+        <p>Our teacher training advisers are here to support those returning to teach maths, physics or languages.</p>
+        <p>
+          We can still offer you support, if you want to teach something else we can <%= link_to_git_mailing_list("send you information via email") %> or you can get advice by <%= link_to_git_mailing_list("attending a return to teaching event") %>.
+        </p>
+    </div>
+  </div>
+</div>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -157,7 +157,7 @@ en:
         teacher_training_adviser/steps/subject_like_to_teach:
           attributes:
             preferred_teaching_subject_id:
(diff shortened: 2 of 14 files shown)


config/locales/en.yml
```diff
@@ -157,7 +157,7 @@ en:
         teacher_training_adviser/steps/subject_like_to_teach:
           attributes:
             preferred_teaching_subject_id:
-              invalid_type: "Please select maths, physics or modern foreign language"
+              inclusion: "Please select maths, physics, modern foreign language or other"
```
peteryates, line 160: Perhaps "Please select a subject" would be easier to grasp here? Or "Please tell us which subject you'd like to teach". Thoughts @croftla?

app/views/teacher_training_adviser/steps/_subject_not_found.html.erb
```diff
@@ -0,0 +1,6 @@
+<h1 class="govuk-heading-l">Get support</h1>
+
+<p>Our teacher training advisers are here to support those returning to teach maths, physics or languages.</p>
+<p>
```
peteryates, line 4: This line is really long in the screenshot. It _might_ be worth wrapping the section in a `.govuk-grid-column-two-thirds`

peteryates (approved):
Code-wise looks great. Couple of minor comments on wording and display

config/locales/en.yml
CroftLA: Choose a subject or other

peteryates (approved):
👍🏽

PR #402, opened 2020-11-23
Release Tag given better name

sfawcett123:
### Trello card
[CD](https://trello.com/c/UQrm3g59/516-platform-continuous-deployment)
### Context
The release name needs to tie in with the PR

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -205,13 +205,14 @@ jobs:
          id: tag_version
          run: |
            MATCH="Merge pull request #[0-9]+"
-           STRING=$(curl -s -X GET "https://api.github.com/repos/${{github.repository}}/commits/${{github.event.inputs.sha}}" | jq ".commit.message")
+           STRING=$(curl -s -X GET "https://api.github.com/repos/${{github.repository}}/commits/${{github.event.inputs.sha}}" | jq -r  ".commit.message")
 
            if [[ ${STRING}  =~ ${MATCH} ]] ;
            then
               echo "Pull request merge"
               TAG=$( echo ${STRING} | grep -oh "[0-9]\+" | head -1 )
-              BODY=$(curl -s -X GET "https://api.github.com/repos/${REPO}/pulls/${TAG}" | jq ".body")
+              echo "PR Number ${TAG}"
+              BODY=$(curl -s -X GET "https://api.github.com/repos/${REPO}/pulls/${TAG}" | jq -r ".body")
            else
               echo "Not a pull request merge"
               TAG="0"
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -194,7 +194,7 @@ jobs:
        - name: Post PR comment
          if: github.event.inputs.environment == 'Review'
          run: |
-          curl --silent 
+          curl --silent  \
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -183,29 +185,43 @@ jobs:
...
-       - name: Bump version and push tag
+       - name: Generate Tag from PR Number
          if: github.ref == 'refs/heads/master' && github.event.inputs.environment == 'Development'
          id: tag_version
-         uses: mathieudutour/github-tag-action@v5
-         with:
-           github_token: ${{ secrets.GITHUB_TOKEN }}
+         run: |
+           MATCH="Merge pull request #[0-9]+"
+           STRING=$(curl -s -X GET "https://api.github.com/repos/${{github.repository}}/commits/${{github.event.inputs.sha}}" | jq ".commit.message")
```
jebw, line 208: Doesn't this need to be `jq -r ".commit.message"` - to strip quotes out from what comes back?

.github/workflows/deploy.yml
```diff
@@ -183,29 +185,43 @@ jobs:
...
+              echo "Pull request merge"
+              TAG=$( echo ${STRING} | grep -oh "[0-9]\+" | head -1 )
+              BODY=$(curl -s -X GET "https://api.github.com/repos/${REPO}/pulls/${TAG}" | jq ".body")
+           else
+              echo "Not a pull request merge"
+              TAG="0"
+              BODY="Nothing"
+           fi
+           echo ::set-output name=new_tag::"PR-${TAG}"
+           echo ::set-output name=changelog::"${BODY}"
```
jebw, line 221: Might be useful to add a step echo'ing out the outputs - could be helpful for future debugging?

.github/workflows/deploy.yml
```diff
@@ -183,29 +185,43 @@ jobs:
...
-           github_token: ${{ secrets.GITHUB_TOKEN }}
+         run: |
+           MATCH="Merge pull request #[0-9]+"
+           STRING=$(curl -s -X GET "https://api.github.com/repos/${{github.repository}}/commits/${{github.event.inputs.sha}}" | jq ".commit.message")
+
+           if [[ ${STRING}  =~ ${MATCH} ]] ;
+           then
+              echo "Pull request merge"
+              TAG=$( echo ${STRING} | grep -oh "[0-9]\+" | head -1 )
+              BODY=$(curl -s -X GET "https://api.github.com/repos/${REPO}/pulls/${TAG}" | jq ".body")
```
jebw, line 214: Likewise - doesn'it it need `jq -r` to strip the quotes?

jebw (changes requested):
Looks good, but from my reading it needs tweaks to the jq lines? Not actually seeing this run in order to test though?

.github/workflows/deploy.yml
sfawcett123: Thanks, I will add that

.github/workflows/deploy.yml
sfawcett123: added

.github/workflows/deploy.yml
sfawcett123: will output TAG the body could be large

jebw (approved):
LGTM

PR #408, opened 2020-11-25
Automate Adviser Deployment to Production

sfawcett123:
### Trello card
[CI CD](https://trello.com/c/UQrm3g59/516-platform-continuous-deployment)
### Context
Implement a process flow :
Created PR -> Review -> :octocat: -> Merge -> Build -> QA (Test) -> Production
 
- Stopped Draft PRs from triggering workflow
- Create draft release if delivery to Development has worked
- Carry out Cypress Tests during QA Stage
- If QA passes change draft release to live release
- If QA Passes deploy to Production

### Guidance to review
Still a work in progress, this phase needs to be working and developers happy with before proceeeding


.github/workflows/deploy.yml
```diff
@@ -14,6 +14,13 @@ on:
       title:
         description: 'Pull Request Title'
         required: false
+      release_tag:
+        description: 'Name of Release Tag'
+        required: false
+      stop:
```
jebw, line 20: this should probably be `stop_after_release` just for future clarity

.github/workflows/deploy.yml
```diff
@@ -221,8 +250,8 @@ jobs:
            echo ::set-output name=new_tag::"PR-${TAG}"
            echo ::set-output name=changelog::"${BODY}"
 
-       - name: Create a GitHub release
-         if: github.ref == 'refs/heads/master' && github.event.inputs.environment == 'Development' && steps.tag_version.outputs.new_tag != '0'
+       - name: Create a GitHub Pre-Release
```
jebw, line 253: Looking lower down this is creating a draft release I think, rather than a 'pre-release' release ? Probably should be renamed accordingly?

.github/workflows/production.yml
```diff
@@ -1,112 +0,0 @@
----
```
jebw, line 1: I don't think this release process (or the release to QA) should be removed until we've actually done a few releases through to production with the new system.

This can get removed in a future PR

.github/workflows/deploy.yml
```diff
@@ -231,22 +260,56 @@ jobs:
...
+         uses: actions/checkout@v2
+         with:
+             repository: DFE-Digital/get-into-teaching-frontend-tests
 
-       - name: Slack Notification
-         if: failure() 
-         uses: rtCamp/action-slack-notify@master
+       - name: Run the Tests for Adviser
+         run: |
+             docker run -i --shm-size=1g   -e CYPRESS_HTTPAUTH_PASSWORD -e CYPRESS_HTTPAUTH_USERNAME  -v $PWD:/test -w /test dfedigital/dfecypress:latest \
```
jebw, line 286: My reading of this is that the Cypress tests will become a blocker to release?

One thing to be aware of is this will mean we always need to do manual releases between a 'breaking release' and sushant getting the cypress tests passing again. If thats by design than that should be fine.

I wonder if we want the cypress test suite change to automatically kick off the deploy to QA? So that it can than carry on into prod when the test suite gets fixed?

jebw (changes requested):
Minor comments on naming, plus I think the prod workflow should stay for the time being

.github/workflows/deploy.yml
sfawcett123: Agreed

.github/workflows/deploy.yml
sfawcett123: Agreed

.github/workflows/deploy.yml
sfawcett123: I don't know how cypress would know what it is releasing.
The manual release I have put in, will deliver to production regardless. So we can decide if it is the test thats failing and still just push through.

Another option, would be to make the Cypress tests not be a blocker and just raise a Slack Message if they fail, though allow the pipeline to go foraward

.github/workflows/production.yml
sfawcett123: Manual workflow is there for that reason, it can release to Test or Production ad-hoc

PR #413, opened 2020-11-27
Don't default preferred education phase for returners

ethax-ross:
By defaulting this in the app we have an edge case where a candidate proceeds as a returner then backs out and starts the new candidate journey. If they do this we will pre-fill the secondary education phase (as its defaulted when they selected returner).

Instead, we will default this value in the API so there is no need to set it for returners at all in the app.


peteryates (approved):
Makes sense

PR #432, opened 2020-11-30
Common Environment Variable Names

sfawcett123:
Used common names for environment variables, and consistently used them
to replace hard coded names, thus allowing for easier code reuse.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -210,7 +210,7 @@ jobs:
                --header "Accept: application/vnd.github.v3+json" \
                --header "Authorization: Bearer ${{ github.token }}" \
                --request POST \
-               --data '{"body": "Review app deployed to ${{needs.deploy.outputs.url}}"}'  \
+               --data '{"body": "Review app deployed to https://${{env.REVIEW_APPLICATION}}-${{github.event.number}}.${{env.DOMAIN}} "}'  \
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -6,9 +6,12 @@ on:
     branches:
       - master
 env:
-  APP_REPOSITORY:  dfedigital/get-teacher-training-adviser-service
-  PROJECT:         Teacher Training Adviser Service
-  DOMAIN:          london.cloudapps.digital
+  DOCKERHUB_REPOSITORY: dfedigital/get-teacher-training-adviser-service
+  DOMAIN:               london.cloudapps.digital
+  PROJECT:              Teacher Training Adviser Service
```
saliceti, line 11: Project has a very specific meaning... Can we use something like `application`, `app`, `app_name`, `service`...

.github/workflows/build.yml
```diff
@@ -6,9 +6,12 @@ on:
...
+  REVIEW_APPLICATION:   review-teacher-training-adviser
+  APPLICATION_NAME:     get-teacher-training-adviser-service
```
saliceti, line 13: If you rename the above, this could become `paas_application_name` or `paas_app_name`

.github/workflows/build.yml
sfawcett123: Renamed to APPLICATION

saliceti (approved):

PR #460, opened 2020-12-14
Inform Slack of Release

sfawcett123:
### Context
When a new release goes out to Production, we should tag the notes attached to that Release and publish them to the slack channel to publicise the release


jebw (approved):
LGTM, approving because its reflecting the release info correctly

_but_ we are currently losing the title of the merge request at present - the body gets copied over to the Release, but not the title? I'm not sure how vital that is if developers are aware of it but if its trivial to get the title included as well I think that would be good

PR #461, opened 2020-12-15
Replace TypeEntity with PickListItem/LookupItem

ethax-ross:
### Trello card

[Trello-358](https://trello.com/c/K6lPR3ev/358-return-types-with-integer-ids-as-integer-rather-than-string)

### Context

The /types endpoints are being deprecated and have been split up into two new endpoints:

/lookup_items (types with a string id)
/pick_list_items (types with an integer id)

We need to replace all calls to the `TypesApi` with the replacement APIs.

- Replace TypeEntity with PickListItem/LookupItem
- Update VCR casettes to mirror new endpoints

### Guidance to review

I've given this a decent amount of manual testing and I've also ran the integration tests against it, all coming back green.


peteryates (approved):
Looks good.

PR #467, opened 2020-12-22
Prevent sign up completion when a 'dead end' step was encountered

ethax-ross:
### Sentry issue

https://sentry.io/organizations/dfe-bat/issues/1864908606/?project=5276960&query=is%3Aunresolved

### Context

Currently it is possible for a candidate to hit a 'dead end' and then manually skip to the review answers screen (which they may have open in another tab) and attempt to complete the sign up, which will result in a 500 error.

This is possible because we currently only check that all 'active' (non-skipped) steps are _valid_, but we neglect to check if they are a 'dead end' step themselves. An example of this would be:

- Returning candidate selects 'Other' as the subject they would like to teach, which is stored as `-1` in the `preferred_teaching_subject_id` attribute (and valid, otherwise they would not to the 'Get support' step).
- Candidate is presented with the 'Get support' dead-end step.
- Candidate manually skips ahead to the review answers screen, is prompted to fill in any other steps they have skipped and can complete the sign up.
- `-1` is submitted to the API for `preferred_teaching_subject_id` and an error is raised as this is not a valid GUID.

To prevent this, we introduce another check on completion to ensure the candidate was able to proceed through all valid steps. If this is not the case the candidate gets redirected to the dead-end step, explaining why they can't proceed (at which point they can choose to go back and change their answer in order to continue).

- Prevent completing sign up on skipping a 'dead end' step

### Guidance to review

I'm not happy with my choice of naming for 'steps that cannot be proceeded through' - I've referenced these as `non_proceedable` steps, but I'm pretty sure that's not a word. I was going to go with 'dead_end' steps, but I wasn't wild about that either as an opposite to `can_proceed?`. Suggestions welcome 😄

### DIFF
--- a/app/controllers/concerns/wizard_steps.rb
+++ b/app/controllers/concerns/wizard_steps.rb
@@ -54,6 +54,8 @@ private
   def next_step_path
     if (next_key = @wizard.next_key)
       step_path next_key
+    elsif (non_proceedable_step = @wizard.first_non_proceedable_step)
+      step_path non_proceedable_step
     elsif (invalid_step = @wizard.first_invalid_step)
       step_path invalid_step
     else # all steps valid so completed
--- a/app/controllers/concerns/wizard_steps.rb
+++ b/app/controllers/concerns/wizard_steps.rb
@@ -54,8 +54,8 @@ private
   def next_step_path
     if (next_key = @wizard.next_key)
       step_path next_key
-    elsif (non_proceedable_step = @wizard.first_non_proceedable_step)
-      step_path non_proceedable_step
+    elsif (exit_step = @wizard.first_exit_step)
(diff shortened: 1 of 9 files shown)


peteryates (approved):
Looks great. I think it _might_ be a bit clearer if the `#valid?` method remains (or is aliased) but have no strong feeling on it.

ethax-ross:
Yeah I think you're right - the reason I renamed it was because `can_proceed` is aliased to the _current step_ so I had to make the new method different so I figured I'd make them both consistent and prefixed with `actrive_steps_` but I think you're right and I should instead rename `can_proceed` to something else, along the lines of `can_proceed_from_current_step?`. I'll do this tomorrow 👍

PR #482, opened 2021-01-12
Add Fail2ban to Rack::Attack config

jebw:
### Trello card

https://trello.com/c/CvoMhTUe

### Context

Bots poking around the site should be banned

1. Added Fail2Ban configuration - this is configured with an environment variable `FAIL2BAN` which is an integer of the minutes to ban an ip for. The rule automatically bans any IP accessing a question able endpoint - ie wordpress or php files or etc/passwd
2. Enable FAIL2BAN for 1 minute in `rolling` for quick testing (eg the Review URLs
3. Enable FAIL2BAN for 3 minutes in `preprod` / aka staging
4. Enabled Rack::Attack in all environments, instead disabling only the req limits in `preprod` / aka staging - to allow for the cypress tests
5. Log to Sentry when an IP is blocked and for how long

### Guidance to review
This PR does not enable the checks in production yet

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -39,8 +39,17 @@ module Rack
 
       blocklist("block hostile bots") do |req|
         Fail2Ban.filter("hostile-bots-#{req.ip}", maxretry: 0, findtime: 1.second, bantime: FAIL2BAN_TIME) do
-          FAIL2BAN_REGEX.match?(CGI.unescape(req.query_string)) ||
+          (
+            FAIL2BAN_REGEX.match?(CGI.unescape(req.query_string)) ||
             FAIL2BAN_REGEX.match?(req.path)
+          ).tap do |should_ban|
+            if should_ban
+              Raven.capture_message <<~BAN_MESSAGE
+                Banning IP: #{req.ip} for #{FAIL2BAN_TIME.to_i / 60} minutes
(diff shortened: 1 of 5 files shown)


ethax-ross (approved):
Looks good 🤖

config/initializers/rack_attack.rb
```diff
@@ -1,25 +1,57 @@
...
+
+      blocklist("block hostile bots") do |req|
+        Fail2Ban.filter("hostile-bots-#{req.ip}", maxretry: 0, findtime: 1.second, bantime: FAIL2BAN_TIME) do
+          (
+            FAIL2BAN_REGEX.match?(CGI.unescape(req.query_string)) ||
+            FAIL2BAN_REGEX.match?(req.path)
+          ).tap do |should_ban|
+            if should_ban
+              Raven.capture_message <<~BAN_MESSAGE
+                Banning IP: #{req.ip} for #{FAIL2BAN_TIME.to_i / 60} minutes
```
ethax-ross, line 48: I think we need to drop the IP address here as it's personally identifiable?

jebw: I've changed it to obscure some of the IP since we may need to verify that the 'right' things are getting banned

ethax-ross (approved):

MylesJarvis:
@jebw can we merge this down now?

PR #490, opened 2021-01-18
Change formbuilder default legend tag to h1

peteryates:
### Trello card

N/A (Accessibility check feedback)

### Context

The form builder wraps fieldsets in only a `<legend>` tag by default, to force it to render a `<h1>` either the `tag:` param must be set or the default configured via an initializer.

Add a default `h1` tag to for legends in a new formbuilder initializer and update specs accordingly. Could possibly be tidied further by adding a custom `have_header` matcher.

### Guidance to review

Are there any pages with more than one fieldset? I couldn't find any at a cursory glance.

### DIFF
--- a/config/initializers/formbuilder.rb
+++ b/config/initializers/formbuilder.rb
@@ -1,0 +1,3 @@
+GOVUKDesignSystemFormBuilder.configure do |conf|
+  conf.default_legend_tag = "h1"
+end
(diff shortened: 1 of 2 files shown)


MylesJarvis (approved):
Looks fine to me.

ethax-ross (approved):
Looks good, I don't think there are any with multiple field sets; each step is only a single question/couple of related fields

PR #492, opened 2021-01-19
reintroduce statuscake on terraform

sfawcett123:
Now that the old statuscake instances have been cleaned up it is possible to restart the new instances


jebw (approved):
Approving but same considerations apply as to the GiT site?

PR #493, opened 2021-01-19
add turnstyle to main workflow

sfawcett123:
### Trello card
[Race Conditions](https://trello.com/c/zFSnSfND/733-race-condition-in-pipelines)

### Context
When two builds are run at the same time, a race condition can occur

Add turnstyle to try and limit impact of race conditions

### Guidance to review
None really

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,9 +7,37 @@ on:
       - master
 
 jobs:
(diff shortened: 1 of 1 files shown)


jebw (approved):
Looks good but I think we're holding off on merges until tomorrow

PR #494, opened 2021-01-20
ignore dependabot labels

sfawcett123:
when Dependabot creates a PR, it still goes to Review ( thats good) but it does not need a Myles review. so prevent the label being added

### Review comment
I tried `if: ! contains(github.event.pull_request.user.login, ‘dependabot’) ` and github hated it so stuck to`if: contains(github.event.pull_request.user.login, ‘dependabot’) == false`


jebw (approved):
LGTM

PR #497, opened 2021-01-22
Add help text on why we check Date of Birth

MylesJarvis:
Following GDS assessment, we were asked about why we collect DoB information. This help text addresses this query.

Added line: We use this information to confirm you are aged 18 or over.

### Guidance to review
Does this render okay?


peteryates (approved):
Looks good.

peteryates:
Just a follow-up question @MylesJarvis, would a checkbox work here?

PR #498, opened 2021-01-22
Update _identity.html.erb

MylesJarvis:
Following the GDS assessment, it was flagged to us that Surname should not be used, but rather the field should be Last name. See here (https://design-system.service.gov.uk/patterns/names/) for guidance.
### Guidance to review
Will this render correctly based on my changes?


MylesJarvis:
Handled in https://github.com/DFE-Digital/get-teacher-training-adviser-service/pull/502

PR #503, opened 2021-01-26
Don't create review apps for dependabot branches

jebw:
### Context

We don't use the Review Apps to verify dependabot branches so we don't need to create them

1. Don't create review apps if branch name starts with 'dependabot/'

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -186,7 +186,7 @@ jobs:
   review:
     name: Review Deployment Process
     needs: build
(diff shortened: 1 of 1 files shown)


sfawcett123 (approved):
I suggest using `contains(github.event.pull_request.user.login, 'dependabot') == false`  as the condition.

jebw:
Sorry should have marked this as WIP - I was still figuring out which syntax GitHub accepted - cheers

MylesJarvis (approved):

PR #521, opened 2021-02-05
Update completed.html.erb

MylesJarvis:
### Trello card
https://trello.com/c/SLCvkKyf/600-content-spike-what-signposting-can-we-give-to-the-tta-confirmation-page-to-help-people-do-their-own-research-eg-go-and-look-at-p
### Context
We had feedback from TTAs that some users had not done too much research before their initial conversation (email/phone etc.). 

Added an unordered list to the confirmation page with various links to key pages users may find useful.

### Guidance to review
Test links and check code is fine. Can we use shortened versions of the links on the TTA service to go to a PB site?

Review Link: https://review-teacher-training-adviser-521.london.cloudapps.digital/teacher_training_adviser/sign_up/completed

### DIFF
--- a/app/views/teacher_training_adviser/steps/completed.html.erb
+++ b/app/views/teacher_training_adviser/steps/completed.html.erb
@@ -15,6 +15,17 @@
         <p class="govuk-body">
           You must respond to the email within 6 weeks of receiving it otherwise you will have to sign up again.  
         </p>
+
+        <p class="govuk-body">
+        While you wait for your adviser, you can: 
+      <ul>
+        <li>Ask any immediate questions using the Get Into Teaching <a href="https://beta-getintoteaching.education.gov.uk/#talk-to-us">online chat</a></li>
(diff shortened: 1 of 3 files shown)


app/views/teacher_training_adviser/steps/completed.html.erb
```diff
@@ -15,6 +15,17 @@
         <p class="govuk-body">
           You must respond to the email within 6 weeks of receiving it otherwise you will have to sign up again.  
         </p>
+
+        <p class="govuk-body">
+        While you wait for your adviser, you can: 
+      <ul>
+        <li>Ask any immediate questions using the Get Into Teaching <a href="https://beta-getintoteaching.education.gov.uk/#talk-to-us">online chat</a></li>
```
ethax-ross, line 22: @MylesJarvis we need to avoid putting hard links to the beta domain in the TTA otherwise they'll break when we transition across and deprecate the service.

This will need a new Rails helper, I'll pick it up once I've finished something I'm in the middle of and update your PR.

peteryates (approved):

PR #530, opened 2021-02-08
Stimulus 2 upgrade

jebw:
### Trello card

https://trello.com/c/teYsKRe4

### Context

We should use the current version of Stimulus

1. Upgraded dependency to Stimulus 2
2. Updated controllers to take advantage of new 'target' and Value api's in Stimulus


ethax-ross (approved):
LGTM

PR #542, opened 2021-02-15
Enabled Fail2Ban in production - initially just for a minute at a time

jebw:
### Trello card

https://trello.com/c/CvoMhTUe

### Context

We should prevent bots poking around the site - particularly if they're looking for 'scammy' files

1. Minimal 1 minute ban

### Guidance to review

Should be safe but we'll have to watch to see what the effects are


ethax-ross (approved):
🤖🔫

jebw:
After further testing it is banning individual CDN nodes - grr - closing till issue is resolved

PR #543, opened 2021-02-15
Added debug data to ban message

jebw:
### Trello card

https://trello.com/c/CvoMhTUe

### Context

We should be banning hostile actors, _but_ also not banning our own CDN if the site is accessed via the CDN

1. Log some debug data in a ban to figure out whats going on. I think I know but want to check before trying to solve it.

### Guidance to review

Not suitable for production but not enabled in production yet anyway

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -8,7 +8,7 @@ module Rack
       .php
     ].freeze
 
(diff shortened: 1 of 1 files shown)


peteryates (approved):
Seems sensible.

ethax-ross (approved):

PR #544, opened 2021-02-16
Bring authentication inline with the app

ethax-ross:
The app and TTA services share much of the base `Wizard`/`Step`/`Store` logic, but its currently duplicated between the two services (we should pull this out into a gem at some point).

This PR updates the TTA service to bring it inline with recent changes to the app (specifically, https://github.com/DFE-Digital/get-into-teaching-app/pull/877/commits/22133ec92b1fc4147e237d4b3f1eb54dac31b9d5, https://github.com/DFE-Digital/get-into-teaching-app/pull/879/commits/cb85be5cd5d4379cfc049d066ace903498d57bc1 and upcoming changes in https://github.com/DFE-Digital/get-into-teaching-app/pull/898)

- Always export matchback attributes

Previously the `Authenticate` step was responsible for exporting the matchback attributes. This made sense as that step was pre-populating the store with them, however going forward the access token logic will be shifting to the Wizard, so it makes sense to tack them on to the export there instead.

- Shift access token requests into Wizard::Base

It makes sense to approach this consistently with the app service and drop the custom Authenticate step subclasses in favour of having the token exchange logic in the Wizard::Base subclasses.

There is also a small change in functionality in this commit; previously the store was purged before issuing a verification code only if a successful authentication had happened in this candidates sign up journey (so if they matched back then went to start the process again).

Now we always purge the store prior to issuing a new verification code. The main reason for this is to avoid potentially complicated sign up scenarios now that users can also authenticate with a magic link.

The only detrimental effect of this change is if a candidate goes through the flow, then back to the first step and doesn't match back; they will need to re-enter any details they added on their first pass through the sign up. This feels like enough of an edge case that its acceptable, given how much simpler and robust the code is if we just clear everything out and treat it as a new sign up (if you go back and change your name/email then the likelihood is its a new/different person signing up anyway).

- Simplify Authenticate step with access_token_used? helper

The `Wizard::Base` now contains a method to determine if/what kind of authentication method has been used.

Instead of manually tracking this information in the `Authenticate` step we can remove the logic and call out to the wizard.

Ensure `timed_one_time_password` dirty tracking is cleared on `ApiError` so that if a user enters a bad code we only try and validate it with the API once (unless they change the code and try again).


peteryates (approved):
This looks fine to me overall.

PR #545, opened 2021-02-16
Hide pre-filled optional fields and steps

ethax-ross:
This is a lift and shift of the app PR https://github.com/DFE-Digital/get-into-teaching-app/pull/907

- Add crm data support to store

We want to hide steps that are optional and have been pre-filled with candidate information.

In order to determine if a step can be pre-filled from the CRM we need to retain the CRM values in the store so that they
can later be queried independently of the data written by the app.

The `fetch` method has also been updated to return values for all keys passed in; this makes the logic to determine if all
values are present in a store simpler (otherwise we would also have to check the `count` to ensure all values are present):

```
fetch(%[...], source: :crm).values.all(&:present?)
```

- Add support for optional steps

An optional step will be skipped if (and only if) the step attributes were all pre-filled from the CRM data.

Data will be send to the API for skipped steps going forward.

This is going to be useful because skipped step values will be `nil` for new candidates and will contain the data already in the CRM for existing candidates. Ultimately this will make it easier to update the API to support sending `nil` values to the CRM (it avoids having to duplicate the logic for 'was this step shown to the user' in the API).

- Give precedence to shown steps during export

We have a number of steps that write to the same attribute. With the way we currently export the later of the steps value is used (regardless of if its skipped or not). This can lead to incorrect data being exported; instead, we need to prioritise the shown steps so that they overwrite data exported from skipped steps.

- Skip optional steps

Updates the optional steps to return true for `optional?`. Updates the `skipped?` logic of these steps to ensure the default 'skip if optional and pre-filled' logic is applied.

Add feature spec to cover skipping optional steps after a matchback.

### DIFF
--- a/app/controllers/teacher_training_adviser/steps_controller.rb
+++ b/app/controllers/teacher_training_adviser/steps_controller.rb
@@ -22,11 +22,15 @@ module TeacherTrainingAdviser
     helper_method :step_path
 
     def wizard_store
-      ::Wizard::Store.new session_store
+      ::Wizard::Store.new app_store, crm_store
     end
 
-    def session_store
+    def app_store
       session[:sign_up] ||= {}
+    end
+
+    def crm_store
+      session[:sign_up_crm] ||= {}
     end
 
     def set_page_title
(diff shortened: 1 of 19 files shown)


peteryates (approved):
Looks good

PR #547, opened 2021-02-18
Add QTS to the list of reqs on the homepage

peteryates:
### Trello card

N/A

### Context and changes

This suggestion was made by the returners team. The PR was raised so we
have a place to discuss the wording rather than to suggest this should
be committed immediately as-is.


MylesJarvis (approved):
LGTM

PR #548, opened 2021-02-18
Left github secret in

sfawcett123:
### Issue
SONAR-TOKEN is trying to be retrieved from GITHUB, but it no longer exists there. this causes the Sonarcube analysis to fail, stopping the pipeline

### Solution
Retrieve SONAR-TOKEN from Azure

### Review
This does not impact the application


peteryates (approved):
👍🏽

PR #552, opened 2021-02-24
Explicitly send the candidate type_id on sign up

ethax-ross:
### Trello card

[Trello-1343](https://trello.com/c/FLYEKcpU/1343-improve-how-we-check-for-returning-teachers)

### Context

We currently do not send to the API if the candidate is a returning teacher vs interested in teaching. This means the logic to determine if a candidate is a returner in the API is a bit convoluted. The API now supports sending the candidate type explicitly, which will mean we can clean that logic up and also has the benefit of being able to pre-fill the "Are you returning to teaching?" step in the TTA service.

- Update API client to include typd_id

- Update ReturningTeacher step to send type_id

In the past we haven't sent a specific attribute value in response to the `ReturningTeacher` step (the previous `returning_to_teaching` attribute was transient and only used to control the flow of the following wizard steps).

The API now accepts a `type_id` as part of the `TeacherTrainingAdviserSignUp` request model, so we will populate this on the `ReturningTeacher` step.

The existing `returning_to_teaching` attribute has been removed but a method that behaves the same takes its place, avoiding further refactoring (really it should be `returning_to_teaching?` but this PR is big enough so we can pick up that refactor later on).

- Remove degree_options from ReturningTeacher step

I think this is a layover from before we refactored the TTA service to use the wizard code in the GiT app. I can't see it referenced anywhere, so removing.

- Removes unused PRIMARY_SUBJECT_ID constant

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,11 +1,11 @@
 GIT
   remote: https://github.com/DFE-Digital/get-into-teaching-api-ruby-client.git
-  revision: 6c2969f0685d49b02527d2d499431498938c3a08
+  revision: fa632dc47559c4669328f0958d6838b64ad684ad
   specs:
-    get_into_teaching_api_client (1.1.10)
+    get_into_teaching_api_client (1.1.12)
       json (~> 2.1, >= 2.1.0)
       typhoeus (~> 1.0, >= 1.0.1)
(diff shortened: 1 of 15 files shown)


peteryates (approved):
Looks good to me

PR #556, opened 2021-03-03
Validate Azure Keys

sfawcett123:
### [Validate Keys ](https://trello.com/c/YTWSmEQL/1341-check-secrets-before-deploy)

Validate the azure keys before usage, to prevent badly formed keys from causing issues further down the pipeline.

### Review
This process does not effect the TTA, it is a step that prevents deployment if the key was invalid. The deployment probably would have failed anyway due to the invalid key, but under certain circumstances the error with the key details could get logged.

### Tested
https://github.com/DFE-Digital/get-teacher-training-adviser-service/actions/runs/617232479

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -13,10 +13,31 @@ on:
         required: false
 
 jobs:
+  validate:
+    name: Validate Secrets
+    runs-on: ubuntu-latest
+    environment:
+       name: ${{ github.event.inputs.environment }}
+    steps:
+       - name: Checkout
+         uses: actions/checkout@v2
+
+       - uses: azure/login@v1
+         with:
+           creds: ${{ secrets.AZURE_CREDENTIALS }}
+
+       - name: Validate Key Vault Secrets
+         uses: DFE-Digital/github-actions/validate-key-vault-secrets@master
+         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -13,10 +13,31 @@ on:
         required: false
 
 jobs:
+  validate:
+    name: Validate Secrets
+    runs-on: ubuntu-latest
+    environment:
+       name: ${{ github.event.inputs.environment }}
+    steps:
+       - name: Checkout
```
vigneshmsft, line 22: checkout is not required for this job?

.github/workflows/deploy.yml
```diff
@@ -13,10 +13,31 @@ on:
...
+       - uses: azure/login@v1
+         with:
+           creds: ${{ secrets.AZURE_CREDENTIALS }}
+
+       - name: Validate Key Vault Secrets
+         uses: DFE-Digital/github-actions/validate-key-vault-secrets@master
+         with:
+           KEY_VAULT: ${{ secrets.KEY_VAULT }}
+           SECRETS: |
+             TTA-KEYS
```
vigneshmsft, line 34:
```suggestion
           SECRETS: TTA-KEYS
```

vigneshmsft (approved):

PR #557, opened 2021-03-03
Show telephone field on review answers step

ethax-ross:
### Trello card

[Trello-1369](https://trello.com/c/ZxF0ggqO/1369-telephone-not-visible-in-review-answers-page-when-equivalent-uk-based-candidate-books-callback)

### Context

The telephone field when filled in as part of the `OverseasTimezone` step was not presented under the personal details section; its now correctly located.

The `UkCallback` step has a mixture of personal/non-personal details and was only showing the non-personal (callback date/time) - it has been updated to render the telephone as well and the answers have been shifted into the personal details section.

Going forward it would make sense to split telephone/callback details into their own section perhaps (instead of trying to split a single steps answers across multiple sections when it contains a mixture of personal/non-personal details).

- Show telephone field on review answers step

### Guidance to review

Screenshot 2021-03-03 at 11 22 51

### DIFF
--- a/app/models/teacher_training_adviser/steps/overseas_time_zone.rb
+++ b/app/models/teacher_training_adviser/steps/overseas_time_zone.rb
@@ -4,6 +4,10 @@ module TeacherTrainingAdviser::Steps
 
     validates :telephone, telephone: true, presence: true
 
+    def self.contains_personal_details?
+      true
+    end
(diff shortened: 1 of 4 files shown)


peteryates (approved):
Looks good to me.

PR #567, opened 2021-03-11
Fix bug where NoDegree step is shown incorrectly

ethax-ross:
### Trello card

[Trello-1405](https://trello.com/c/Ccwvwt0v/1405-bug-fix-no-degree-step-can-be-incorrectly-shown-on-returner-journey)

### Context

If a candidate initially starts their journey as a non-returner and then selects no degree we will show them the exit step informing them that they have to have a degree.

If they they go back and try to carry on as a returner they will see the same exit step. Instead, we should be ignoring their selected degree option when on the returner journey.

- Fix bug where NoDegree step is shown incorrectly


peteryates (approved):
Makes sense.

PR #572, opened 2021-03-15
Bug fix callback scheduled for returner

ethax-ross:
### Trello card

[Trello-1405](https://trello.com/c/Ccwvwt0v/1405-bug-fix-no-degree-step-can-be-incorrectly-shown-on-returner-journey)

### Context

If a candidate goes all the way through the equivalent degree/UK path and schedules a phone call, then edits the answer for 'are you returning to teaching' and proceeds down the returner path as a UK candidate we end up exporting their `phone_call_scheduled_at` value incorrectly and the request fails the API validation.

Instead, we should treat the `UkCallback` step as `skipped?` unless they are shown the `HaveADegree` step.

- Bug fix callback scheduled for returner


peteryates (approved):
Makes sense

PR #574, opened 2021-03-15
Enable circuit breaker

jkempster34:
### Trello card
https://trello.com/c/e9vBMLJE

### Context
Circuit breaker middleware has been added to the API Client Gem. It is enabled via the following configuration:
```rb
GetIntoTeachingApiClient.configure do |config|
  config.circuit_breaker = {
    enabled: true, # Enables the circuit breaker middleware
    threshold: 5, # Number of failures that will cause the circuit breaker to trip
    timeout: 5.minutes # Amount of time until the circuit breaker will attempt to recover
  }
end
```

- Enable circuit breaker

### DIFF
--- a/config/initializers/get_into_teaching_api_client.rb
+++ b/config/initializers/get_into_teaching_api_client.rb
@@ -8,4 +8,10 @@ GetIntoTeachingApiClient.configure do |config|
   config.api_key["Authorization"] = ENV["GIT_API_TOKEN"].presence || Rails.application.credentials.config[:api_key].presence
 
   config.cache_store = Rails.cache
+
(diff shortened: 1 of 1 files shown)


ethax-ross (approved):
LGTM - I think its worth logging to Sentry when a circuit breaks though so we have a clear indication for when users are going to see a change/service unavailable message. It should just be a case of calling `capture_message` here:

https://github.com/DFE-Digital/get-into-teaching-app/blob/f392f67121a0e8814d66e0de738847ad42921740/app/controllers/concerns/circuit_breaker.rb#L7

PR #589, opened 2021-03-25
Build from working cached build

jebw:
### Trello card

https://trello.com/c/c4lViBfR

### Context

Builds are broken because they depend upon a no longer available mimemagic gem, and the docker cache has expired.

1. Use the last working build as a cache source as a stop gap fix


sfawcett123 (approved):
Looks like it may just work :)

PR #590, opened 2021-03-25
Avoid reinstalling gems on every build

jebw:
Creating the app sha file will result in a new file on every build. Since this
is before the Gem installation, it will result in Gems being reinstalled on
every build irrespective of whether the cache could supply the Gems.

Instead move it to the final step since its guarenteed to be different for every
build.


leoapost (approved):
Looks good

PR #600, opened 2021-03-29
Updated phone number

MylesJarvis:
### Trello card
https://trello.com/c/H20rftsn/1428-change-phone-number-on-site-to-0800-389-2500

### Context
We are updating the phone number back to the old number, as we take over the new domain.

Any phone number reference from 2501 -> 2500.

### Guidance to review
Has every instance been corrected? I used find and then replaced individually, so I imagine so.


ethax-ross (approved):
I think you've got them all, I couldn't find any others

PR #602, opened 2021-03-31
Draft Public Beta cookie banner

MylesJarvis:

### DIFF
--- a/app/views/shared/_cookie_banner.html.erb
+++ b/app/views/shared/_cookie_banner.html.erb
@@ -3,21 +3,11 @@
 
   <div id="cookie-banner-message" class="govuk-width-container"
     data-cookie-acceptance-target="banner">
-
+    <h1>Tell us whether you accept cookies</h1>
     <p>
-        Welcome to the new Get an Adviser service. We invite you to try the
-        latest beta version of the site and help us improve the services we offer.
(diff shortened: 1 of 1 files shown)


app/views/shared/_cookie_banner.html.erb
```diff
@@ -3,21 +3,11 @@
...
-        Welcome to the new Get an Adviser service. We invite you to try the
-        latest beta version of the site and help us improve the services we offer.
-    </p>
+    <p><strong>Tell us whether you accept cookies</strong></p>
     <p>
-        To use the new website, we need you to agree to our cookie policy. This
-        means we can use your data to understand how to improve the website
-        for all of our users.
+    We use <%= link_to "cookies to collect information", cookies_path %> about how you use this site. 
+    We use this information to make the website work as well as possible and improve our services.
```
ethax-ross, line 9: This reads a bit funny to me: "work as well as possible" -- are they not really just the same thing? "to make the website work and improve our services" reads better to me?

MylesJarvis: I've used this for inspiration, so was hoping to keep as is


image

ethax-ross: Fair enough 👍

ethax-ross (approved):

PR #606, opened 2021-04-07
Update GTM service ID for production

ethax-ross:
> :warning: Pending a green light from Shivani

### Trello card

[Trello-1490](https://trello.com/c/1tODW5fa/1490-update-gtm-for-public-beta)

### Context

We are using a new container when going into public beta in order to differentiate the metrics from private beta.

- Update GTM service ID for production


ethax-ross:
Shivani confirmed this is no longer needed.

PR #611, opened 2021-04-09
Redirect beta domain to new prod domain

ethax-ross:
- Redirect beta domain to new prod domain

We now have a non-beta production domain; all requests to the beta domain should be redirected.

- Serve robots.txt from Rails

We want to serve the robots.txt file from Rails so that we can intercept calls on the beta- domain and redirect them.


peteryates (approved):
Looks good to me!

PR #625, opened 2021-04-16
Enable ActiveRecord

ethax-ross:
### Trello card

[Trello-1514](https://trello.com/c/fnHpXkZt/1514-add-postgres-to-tta-service)

### Context

We require a storage mechanism for the TTA feedback form submissions and our current options aren't a good fit for this feature (GA, Prometheus, Logs).

- Enable ActiveRecord

Enable ActiveRecord in Rails and configure production database to point to the `postgres` instance in `VCAP_SERVICES`.

Uses a postgres service in the GitHub Action for running tests.

### Guidance to review

Nothing has changed as far as the user is concerned in this PR; as long as the app passes CI and boots in the review space it should be good.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,7 +44,7 @@ jobs:
           SLACK_MESSAGE: Workflow concurrency issue
           SLACK_TITLE: Turnstyle failure ${{ github.workflow }}
           SLACK_WEBHOOK: ${{ steps.azSecret.outputs.SLACK-WEBHOOK }}
-          
+
   build:
     name: Build and Test
     runs-on: ubuntu-latest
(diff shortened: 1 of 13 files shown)


.github/workflows/build.yml
```diff
@@ -53,8 +53,20 @@ jobs:
     services:
       registry:
         image: registry:2
-        ports: 
+        ports:
           - 5000:5000
+      postgres:
+        image: postgres
```
peteryates, line 59: Worth specifying a version?

peteryates (approved):
Looks fine to me. Left a comment on specifying a postgres version but not sure how important it is here.

.github/workflows/build.yml
ethax-ross: Yeah good shout I think we use v11 I'll get it updated

leoapost (approved):
Looks good.

PR #630, opened 2021-04-20
Add invisible captcha protection to feedback form

ethax-ross:
### Trello card

### Context

At the moment a bot could easily submit the feedback form; we want to prevent that so that we only get authentic submissions from candidates.

Add `invisible_captcha` gem, which puts three forms of protection onto the page:

- A hidden honey pot field (if filled in the submission is rejected).
- A 'spinner' field (if modified the submission is rejected).
- A minimum submission time (too-fast submissions are rejected).

If a submission is rejected we render a HEAD 200 response, which should be enough to make the bot think it was successful/hopefully it then moves on.

- Add invisible captcha protection to feedback form

### Guidance to review

You can manually trigger the bot protection by filling in the honey pot field before submitting.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -51,6 +51,8 @@ gem "sentry-ruby"
 # Ignore cloudfront IPs when getting customer IP address
 gem "actionpack-cloudfront"
 
+gem "invisible_captcha"
+
 group :development, :test do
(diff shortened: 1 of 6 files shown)


leoapost (approved):
Looks good

PR #631, opened 2021-04-20
Support multiple basic auth users

ethax-ross:
### Trello card

### Context

The TTA service currently uses basic auth in production-like environments (that aren't the actual prod env) to restrict access. Going forward, we want to be able to restrict access in the production environment as well and support multiple users that can be scoped to certain actions.

To achieve this, the format of the credential for basic auth has been updated to support multiple users (the users can share the same username, which means we can treat their username as their role/access level):

```
http_auth: user1=pass1,user2=pass2,...
```

The basic auth will be enforced by default in all production-like (but not prod itself) environments and can be overriden on a per-controller basis. This means we can disable the site-wide auth in production apart from in the feedbacks controller, where we could also restrict access by checking the username:

```
before_action :restrict_access, only: %i[index export]

def restrict_access
  render_forbidden unless session[:username] == "feedback"
end

def authenticate?
  true
end
```

- Remove basic auth from healthcheck requests

We don't enforce basic auth on the healthcheck endpoint so this shouldn't be needed.

- Add http_auth credentials to preprod/rolling

We are going to show the site-wide auth on all pages in preprod/rolling; this adds the necessary credentials in the new format.

- Support multiple basic auth credentials

Extract the basic auth logic from the controller into a lib; updates to support multiple logins, as we will need this going forward (for restricting access to the feedback section).

The controller will set a `session[:username]` when auth is successful so that we can later scope certain actions to particular users.

Removes redundant feature spec testing basic auth (we cover this in a request spec already).

- Remove unused basic auth env vars

The and `HTTPAUTH_*` env vars are no longer used; we now have a `http_auth` credential set in the Rails creds.


leoapost (approved):
Looks good to me

PR #632, opened 2021-04-20
Enable viewing and exporting feedback to CSV

ethax-ross:
### Trello card

### Context

We now accept feedback submissions that are persisted into the postgres database. We want to be able to view the most recent submissions and export within a date range to a CSV file.

Access to this feature must be restricted in all environments using a `feedback` username and strong password in production.

- Enable transactional fixtures

Ensure each test runs against a clean database.

- Add human_boolean helper

We are going to be rendering boolean values in a table for displaying recent
feedback; it will be better to output 'Yes' and 'No' instead of `true` and
`false` directly.

- Add query scopes for recent feedback in a date range

When exporting feedback we will allow the user to choose a date range to export from and intend to display the most recent first.

- Add FeedbackSearch model

We want to be able to query feedback around specific dates so that the user can export as CSV.

- Add FeedbackExporter

After querying feedback we want to be able to export it as CSV.

- Add ability to view/export feedback

An `index` action renders the 10 most recent feedback in a table, along with a form allowing a user to export feedback as CSV.

The `export` action queries the feedback within the given time range and exports it.

- Support rendering a forbidden error page

We will need this now that we are accepting basic auth conditionally for some users against certain actions.

- Secure feedback index/export

Add a `feedback` user to the credentials.

Secure the index/export actions so that they can only be accessed by the `feedback` user in all environments.

### Guidance to review

Best reviewed by-commit, I can split it out into several PRs if its too much. Note you have to log in with the new `feedback` user or you will get `403 Forbidden` on the index/export actions.

Screenshot 2021-04-20 at 15 40 06


peteryates (approved):
Nice, very thorough!

PR #633, opened 2021-04-21
Minor follow up tweaks to CSV exporting

ethax-ross:
### Trello card

### Context

A CSV cell that begins with `+`, `-`, `@` or `=` could contain malicious formula. To prevent CSV injection these cells must be prefixed with a `'`.

I noticed that in the review environment a validation error was occurring when the "Created on or before" date defaults to today; I think this is because the current time is being compared in the validation conditional, even though I couldn't replicate it locally. Updating to be `end_of_day` to see if that resolves it.

- Escape CSV injection payloads
- Tweak created_on_or_before validation to be looser
- Prevent flakiness in environment tests

We're seeing flakiness around the tests against `BasicAuth` due to the credentials being memoized in a class variable and not cleared between runs.

Add helper to mock `BasicAuth.authenticate`, which bypasses the memoized credentials.


leoapost (approved):
Looks good to me

PR #639, opened 2021-04-21
Use localisation for form label/legend/hint text

ethax-ross:
### Trello card

[Trello-1518](https://trello.com/c/M0peQNo7/1518-update-tta-service-to-shift-form-text-into-translations-file)

### Context

The GOV.UK form builder will infer the label/legend/hint text from the model/attribute name and load it from the localisations file by default.

We should shift all text where possible to the localisations file (some form elements don't use GOV.UK builder or do not have an associated attribute; these can be left as-is for now).

We should also remove redundant classes for links and p tags; the styling is applied globally so these aren't required.

- Use localisation for form label/legend/hint text

### Guidance to review

I manually verified each of the changes to the steps as I was making the changes to ensure they stayed consistent as far as the user is concerned.


leoapost (approved):
Looks good to me

PR #642, opened 2021-04-26
Isolate API client HTTP cache store

ethax-ross:
### Context

The API client library is being updated to support cache invalidation under certain scenarios (for example, posting a new teaching event needs to clear the event search cache).

As invalidating specific response caches would be reasonably tricky (the caching is performed by a separate middleware) the API client instead blows away its entire cache store (the cache is short-lived and cheap). It is no longer safe to use `Rails.cache` for the API client and instead we need to isolate the cache, so we don't invalidate anything apart from cached HTTP responses.

The TTA service doesn't support any of the endpoints that invalidate cache at the moment, but we should still isolate the cache to protect against future changes.

- Isolate Api client HTTP cache store

Use a new in-memory cache for development and a separate Redis namespace in production.

### Guidance to review

Api client change: https://github.com/DFE-Digital/get-into-teaching-api-ruby-client/pull/49

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -53,6 +53,7 @@ Rails.application.configure do
 
   config.x.git_api_endpoint = "https://get-into-teaching-api-dev.london.cloudapps.digital"
   config.x.enable_beta_redirects = false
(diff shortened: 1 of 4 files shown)


leoapost (approved):
Looks good

PR #646, opened 2021-04-29
Migrate to GHCR

sfawcett123:
### Trello card
https://trello.com/c/F01QXCqc/1526-migrate-to-github-container-registry

### Context
Docker Hub have limited the amount of users and the amount of pulls that we can do, It is also another service we need to manage, when we can now use GHCR which keeps the docker image with the source and does not have the restrictions of docker hub.

When building or deploying docker images store and retrieve them from GitHub Container Repository.

### Guidance to review
No change to application


peteryates (approved):
Looks good to me. Looks like there's a bit of config missing though, as it's still reaching out to Docker's registry.

> Error: Error response from daemon: Get https://registry-1.docker.io/v2/: unauthorized: incorrect username or password

PR #649, opened 2021-04-30
Re-introduce overseas callback booking

ethax-ross:
### Trello card

[Trello-1554](https://trello.com/c/y1MtRspx/1554-re-enable-international-callbacks-in-the-dev-test-environment-only)

### Context

We want to bring back in and updates the original overseas callback booking step, so that if you are a candidate that has an equivalent degree and is based overseas we will prompt you to provide your time zone and then display a list of callback booking quotas in your local time zone to choose from.

We only want to enable this in non-production environments for now so that the CRM team can test the logic in our rolling environment.

- Re-introduce overseas callback booking

### Guidance to review

Due to the way steps are displayed on the review answers page (they either contain personal details or they don't), the timezone is displayed under 'personal details' (as it is asked on the same step as telephone, which is personal). It would be a fair amount of work to allow splitting these out, but less work if we wanted to put telephone/timezone/callback details in their own, separate group in the review answers page.

| Telephone/TimeZone      | Callback Slot | Answers |
| ----------- | ----------- |------------|
| Screenshot 2021-04-30 at 09 16 56 | Screenshot 2021-04-30 at 09 17 03 | Screenshot 2021-04-30 at 09 17 19 |


MylesJarvis (dismissed):
LGTM. @ethax-ross the confirmation page for this specific journey is arguably a bit off now that the next thing they'll receive is a call, not an email (which is in our generic confirmation page). We might need to look into bringing forward the ticket on custom confirmation pages, specifically for this route.

app/views/teacher_training_adviser/steps/_overseas_callback.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= f.govuk_fieldset legend: { text: 'Choose a time' } do %>
+
+<div class="govuk-form-group">
+  <%= f.label :phone_call_scheduled_at, "Select your preferred day and time for a callback", class: "govuk-label" %>
+  <%= f.select :phone_call_scheduled_at, callback_options(f.object.class.callback_booking_quotas), {}, { class: "govuk-select" } %>
```
peteryates, line 5: Doesn't matter too much here, but the latest version of the govuk form builder has a [new `#govuk_select` method](https://govuk-form-builder.netlify.app/form-elements/select/#taking-greater-control-over-the-options) which is a direct wrapper for `#select`

peteryates (dismissed):
Nice

app/views/teacher_training_adviser/steps/_overseas_callback.html.erb
ethax-ross: Oh nice! I'll update that - I tried using the `govuk_collection_select` but it was going to be more work to manually churn out the options in the correct format

peteryates (approved):

PR #666, opened 2021-05-11
Switch to SemanticLogger in Production

ethax-ross:
### Trello card

[Trello-1573](https://trello.com/c/futEUxSe/1573-trim-some-of-the-logging-fat)

### Context

We want to log in a compact JSON format, as we already do in the GiT app.

Add the SemanticLogger gem and configure so it will work well with logit.io.

Log at the `info` level in production instead of `debug`.

- Switch to SemanticLogger


leoapost (approved):
LGTM

PR #685, opened 2021-05-17
Improve email format validation

ethax-ross:
The built-in URI email regex appears to allow emails without fully qualified domains, such as `some@thing`. We use a custom email format validator in the GiT app that is more suitable for our use-case, so this is a lift and shift of that validator into the TTA codebase.


leoapost (approved):
LGTM

MylesJarvis (approved):

PR #687, opened 2021-05-20
help text confusing

EmmaFrith:
help text talks about 'the return to teaching route' - this is super confusing as there's no obvious part of the form (from a user's perspective) that is the 'return to teaching route'

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -152,7 +152,7 @@ en:
       teacher_training_adviser_steps_stage_interested_teaching:
         preferred_education_phase_id: "Select a stage even if you're not sure yet, as this will not affect the sign up process."
       teacher_training_adviser_steps_returning_teacher:
-        type_id: "If you have an overseas teaching qualification that's equivalent to our qualified teacher status (QTS), then select the 'yes' option and take the 'returning to teaching' route."
+        type_id: "Select 'yes' if you have an overseas teaching qualification that's equivalent to qualified teacher status (QTS)."
       teacher_training_adviser_steps_have_a_degree:
         degree_options: "To use this service your bachelor's degree or predicted grade needs to be a 2:2 class honours or above."
       teacher_training_adviser_steps_overseas_telephone:


ethax-ross (approved):

MylesJarvis:
@EmmaFrith Do you think it's worth linking to the following guidance to help users understand equivalence to QTS? 
https://www.gov.uk/guidance/qualified-teacher-status-qts#teachers-from-the-eu-iceland-liechtenstein-norway-switzerland-australia-canada-new-zealand-or-usa

EmmaFrith:
> @EmmaFrith Do you think it's worth linking to the following guidance to help users understand equivalence to QTS?
> https://www.gov.uk/guidance/qualified-teacher-status-qts#teachers-from-the-eu-iceland-liechtenstein-norway-switzerland-australia-canada-new-zealand-or-usa

I actually think the whole 'equivalent to QTS' part of this phrase is pretty confusing - and that link to gov.uk doesn't really clarify it very well. For example, in the section on 'EU, Iceland, Liechtenstein, Norway, Switzerland, Australia, Canada, New Zealand or USA' it just refers to 'a teaching qualification'. 

How about we re-word this sentence to:

Select 'yes' [if you're an overseas qualified teacher](https://getintoteaching.education.gov.uk/international-candidates#already-a-teacher-teach-in-england). 

This link would take them to the relevant section on GIT, which also includes a link to that gov.uk page you mention.

@MylesJarvis

MylesJarvis:
I think the issue might be that in an ideal world we don't want those with non-QTS equivalents going through the form and getting to an adviser, as it would be a waste of resources. With that in mind, I think we need to keep QTS in there. @EmmaFrith

EmmaFrith:
> I think the issue might be that in an ideal world we don't want those with non-QTS equivalents going through the form and getting to an adviser, as it would be a waste of resources. With that in mind, I think we need to keep QTS in there. @EmmaFrith

The problem is that there's no where clear and concise telling you what actually is equivalent to QTS...

PR #693, opened 2021-05-21
put link to privacy policy in the text rather than separate

EmmaFrith:
@ethax-ross - I'm guessing I've done something funky here!? was unable to review as I made the change

### DIFF
--- a/app/views/teacher_training_adviser/steps/_accept_privacy_policy.html.erb
+++ b/app/views/teacher_training_adviser/steps/_accept_privacy_policy.html.erb
@@ -1,7 +1,7 @@
 <% policy_id = session[:privacy_policy_id].presence || GetIntoTeachingApiClient::PrivacyPoliciesApi.new.get_latest_privacy_policy.id %>
 
 <%= f.govuk_check_boxes_fieldset :accepted_policy_id, legend: { text: "Read and accept the privacy policy"} do %>
-  <p>Read our [privacy policy](/privacy-policy) before you continue. It explains how we use your personal data.</p>
+  <p>Read our <%= link_to("privacy policy", privacy_policy_path(id: policy_id), target: "blank") %> before you continue. It explains how we use your personal data.</p>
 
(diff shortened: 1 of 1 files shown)


MylesJarvis:
image

One for @ethax-ross to explain, but my interpretation of the code is that it's set up like it is because we link to a specific privacy policy ID stored in the CRM and then pull that for the private policy page.

MylesJarvis:
And it would need to be coded in html to link effectively

ethax-ross:
I've pushed a commit to fix the link @EmmaFrith, it wasn't working the way you had it because the TTA service doesn't support markdown like the GiT website does I'm afraid

peteryates (dismissed):

app/views/teacher_training_adviser/steps/_accept_privacy_policy.html.erb
```diff
@@ -1,11 +1,7 @@
 <% policy_id = session[:privacy_policy_id].presence || GetIntoTeachingApiClient::PrivacyPoliciesApi.new.get_latest_privacy_policy.id %>
 
 <%= f.govuk_check_boxes_fieldset :accepted_policy_id, legend: { text: "Read and accept the privacy policy"} do %>
-  <p>Our privacy notice explains how we use your personal data. It is important that you have read this notice before continuing.</p>
-
-  <div class="govuk-form-group">
-    <%= link_to "Privacy policy", privacy_policy_path(:id => policy_id) %>
-  </div>
+  <p>Read our <%= link_to("privacy policy", privacy_policy_path, target: "blank") %> before you continue. It explains how we use your personal data.</p>
```
MylesJarvis, line 4: @ethax-ross I note the code in the removals also had a reference to `policy_id`, do we need to reference that here or not?

ethax-ross: Ah, yes well spotted - that ensures the policy they accept is the one that we have shown them (on the off chance it changes after they view it but before they submit the form). Updated.

peteryates (approved):

PR #715, opened 2021-06-07
Fix bug on pre-filling country dial-in code

ethax-ross:
If the user gets to the step asking them for their overseas telephone number before they have specified their country of residence we get an exception due to the `country_name` being `nil`. This should only be possible if someone skips ahead by manually entering the URL of a future step or if their session expires part way through, perhaps.

Default to no country-code being pre-filled if the country of residence is not yet known.


leoapost (approved):

peteryates (approved):
Makes sense to me

PR #717, opened 2021-06-08
Show booking slots as available unless they are in the next hour

ethax-ross:
### Trello card

[Trello-1626](https://trello.com/c/SRdITVG8/1626-international-dialing-outstanding-issues)

### Context

We were previously limiting the earliest booking slots to the following day, however we now want to allow candidates to book a slot on the current day, but omit any in the next hour to give TPUK enough time to receive and process the
booking.

- Show booking slots as available unless they are in the next hour


leoapost (approved):
looks good

PR #754, opened 2021-07-20
Use default Prometheus client data store

ethax-ross:
We don't need to use the `DataFileStore` as we only run a single Puma process.


peteryates (approved):
👍🏽

PR #756, opened 2021-07-22
Add helptext for Year of Start

MylesJarvis:
We're seeing many people click 2021 when actually they want to apply in 2022. Adding some suggested help text to see if this helps.

This is the first of 2 changes. The other (changing drop down options) will be done in a separate PR.


ethax-ross (approved):
Is it worth making this dynamic so it changes according to the current year and if we're after September?

MylesJarvis:
> Is it worth making this dynamic so it changes according to the current year and if we're after September?

It will simply need removing, not amending, at the end of this cycle application window (6 September-ish). I've added a ticket for it.

PR #763, opened 2021-07-29
Allow some candidates to re-subscribe for a TTA

ethax-ross:
> :warning: **Do not merge until CRM team have tested on the review instance**

### Trello card

[Trello-1698](https://trello.com/c/yG3hzAuh/1698-development-unblock-get-an-adviser-applicants-who-have-previously-had-an-adviser)

### Context

Previously we were using the `already_subscribed_to_teacher_training_adviser` field to determine if we prevent the user from continuing with their sign up. Instead, we want to use the `can_subscribe_to_teacher_training_adviser` field which will allow some candidates that have already subscribed to re-subscribe (depending on their adviser status as determined by the
API).

- Bump API Client version
- Allow some candidates to re-sign up for a TTA

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,12 +1,12 @@
 GIT
   remote: https://github.com/DFE-Digital/get-into-teaching-api-ruby-client.git
-  revision: 3ed0a85103fd8ba51b120076314f511d0dcfb4dd
+  revision: 268033a9e88dcbbf8bba1a5deb40f92a243e9c71
   specs:
     get_into_teaching_api_client (1.1.17)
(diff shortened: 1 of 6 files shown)


MylesJarvis (commented):
Going to remove myself on this and allow for a dev review

MylesJarvis:
Never mind I can't as code owner - will await a dev review and then approve after

leoapost (approved):
Looks ok to me

leoapost (approved):

leoapost (dismissed):

leoapost (approved):

PR #785, opened 2021-08-17
Add contract test for candidate re-subscription

ethax-ross:
When a candidate is in a closed state (their `adviserStatusId` is one of a subset that are 'closed') we enable them to re-subscribe for a TTA.

Add contract test capturing the scenario and request made to the API for candidates that are re-subscribing for an adviser.


peteryates (approved):
Looks good to me 👌🏽

PR #797, opened 2021-09-03
Show next year on 7th September instead of 18th

ethax-ross:
The marketing team have asked for us to display the next teacher training year on the 7th September instead of the 18th.


jkempster34 (approved):
👍

PR #831, opened 2021-09-29
Add draft details of why we capture DOB

peteryates:
This is a draft PR that adds some more details to the date of birth page on why it's necessary to capture.


gemmadallmandfe:
Thanks for starting this off @peteryates

On further review, we do link to the privacy notice at the start of the Get an Adviser flow: https://adviser-getintoteaching.education.gov.uk/teacher_training_adviser/sign_up/identity
image

I'd suggest that means we don't need to do so again on the age page?

But we absolutely need to expand on the reasons!

PR #832, opened 2021-10-01
Support custom channel in TTA sign up

ethax-ross:
### Trello card

[Trello-2498](https://trello.com/c/KKq8jc3M/2498-on-campus-channel-attribution-for-tta-sign-up)

### Context

The events team are pushing TTA sign ups at on-campus events and want to be able to attribute a successful sign up to a particular event. The API now accepts a `channe_id` in the TTA sign up request model, so this PR updates the TTA root URL to forward on a `channel` attribute, that will be used for the `channel_id` when signing up. It is persisted as part of the first (identity) step.

To sign up the URL format is:

```
/?channel=<channel_id>
```

- Support custom channel in TTA sign up

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,12 +1,12 @@
 GIT
   remote: https://github.com/DFE-Digital/get-into-teaching-api-ruby-client.git
-  revision: 268033a9e88dcbbf8bba1a5deb40f92a243e9c71
+  revision: 1b33c1debd3914590d4873ec7d817e933304dc64
   specs:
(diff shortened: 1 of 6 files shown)


peteryates (dismissed):
Looks good. Originally had slight reservations around `permit!` but having had a closer look can't really see a nicer way (maybe `params.to_unsafe_h` but it's more or less the same here)

ethax-ross:
I think `to_unsafe_h` is probably more semantic here, I'll push an update 👍

peteryates (approved):
Looks good to me

PR #841, opened 2021-10-13
Add BAM tracking pixel to completion page

ethax-ross:
### Trello card

[Trello-2499](https://trello.com/c/NEH6fGRo/2499-new-bam-pixel-for-on-campus-ways-to-train-sign-ml-ups)

### Context

The TTA service now accepts a custom channel for sign ups so that they can be attributed to in-person events. In order to analyse the data we also need the BAM tracking pixel added; this is a lift and shift from the app and will be replaced with GTM in the future.

- Add BAM tracking pixel to completion page

### Guidance to review

I've enabled in all environments to make it easier to test (there's no harm in it as the pixel is tied to the prod URL so the page views won't count)


peteryates (approved):
Looks good

PR #900, opened 2022-01-14
Update various dependencies

ethax-ross:
Update @rails/ujs, jest, rails_semantic_logger, shoulda-matchers, govuk-frontend, brakeman, sentry-rails

Closes #897, #896, #895, #891, #887, #886, #881

I've left the Rails update to a separate PR.


app/webpacker/styles/application.scss
```diff
@@ -40,6 +40,10 @@ $govuk-global-styles: true;
   background-color: #0076bd;
 }
 
+.govuk-hint  {
```
peteryates, line 43: I don't think this is needed if form builder is also bumped to 3.0.0.

ethax-ross: That worked thanks 👍

peteryates (approved):
Looks good!

PR #946, opened 2022-02-22
Allow review apps to be built on draft PRs

peteryates:
### Context and changes

It makes sense to build review apps for draft PRs.

- Allow review app deployment for draft PRs
- Trim any trailing whitespace

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,10 +8,10 @@ on:
 
 jobs:
   build:
-    name: Build 
+    name: Build
     runs-on: ubuntu-latest
     outputs:
-      DOCKER_IMAGE: ${{ steps.docker.outputs.DOCKER_IMAGE }} 
+      DOCKER_IMAGE: ${{ steps.docker.outputs.DOCKER_IMAGE }}
     steps:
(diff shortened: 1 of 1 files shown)


peteryates:
Recommend [ignoring the whitespace](https://github.com/DFE-Digital/get-teacher-training-adviser-service/pull/946/files?diff=split&w=1) on this one

sfawcett123 (changes requested):
On the line 

    types: [assigned, opened, synchronize, reopened]
    
   you may need to add ready_for_review and some others, if you intend on doind things like this ( see https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) because when you switch the PR from draft to read_for_review it will not trigger a review build, which you may need

peteryates:
@steve - won't this change mean that a review app is generated on every (non-dependabot, non-master) push regardless of whether it's draft or not? Not sure how changing the `types` would have an impact

peteryates:
Ok, I've added it but I'm not totally certain why. Isn't building after every `synchronize` enough? Building every time commits are pushed should have us covered.

PR #949, opened 2022-02-24
Improve network policies

sfawcett123:
Manage Network polices using terraform


terraform/paas/variables.tf
```diff
@@ -43,10 +43,17 @@ variable "paas_linked_services" {
...
+variable "paas_monitoring_space" {
+  default = "sandbox"
+}
+
 variable "paas_space" {
   default = "sandbox"
 }
 
+variable "paas_monitoring_app" {
+  default = "prometheus-dev-get-into-teaching"
```
saliceti, line 55: If we always expect to pass a value, we can remove the default

terraform/paas/variables.tf
```diff
@@ -43,10 +43,17 @@ variable "paas_linked_services" {
   , "get-into-teaching-api-dev-pg-common-svc"]
 }
 
+variable "paas_monitoring_space" {
+  default = "sandbox"
```
saliceti, line 47: If we always expect to pass a value, we can remove the default

terraform/paas/variables.tf
sfawcett123: removed

saliceti (approved):

PR #963, opened 2022-03-08
Fix integration tests not running

ethax-ross:
The TTA service hadn't been setup with selenium so the tests were failing. The docker image is now also bigger as it includes chromium, so I've had to up the disk quota.

The chromium package isn't compatible with a previous security fix - I've removed the workaround for now; it may still get flagged by snky, in which case I'll need to find another way around it.


peteryates (approved):
Looks good

PR #967, opened 2022-03-16
Handle when invalid channel id is specified

ethax-ross:
### Trello card

[Trello-3051](https://trello.com/c/oMCKU9xZ/3051-improve-error-message-when-channel-id-is-invalid)

### Context

Our on-campus events provide a `channel=` query string parameter (by way of a QR code/URL). If the channel is invalid we currently show an unhelpful error message.

Whilst it shouldn't be possible for users to encounter this scenario, if the QR code author makes a mistake it _could_ lead to
the user running into it. 

Handle the error more gracefully by clearing the channel and allowing the user to continue.

In addition to the `channel` we already record we want to allow BAM to send a `sub_channel` in the query string so that they can better attribute sign ups.

- Handle when invalid channel id is specified

Also fixes a bug whereby the channel ID was dropped if the user hits a validation error on the identity step.

- Expose sub_channel for BAM tracking pixel

Store the `sub_channel` if present in the query string and expose it via the DOM so that we can retrieve it in GTM and include it in the magic pixel request.

### DIFF
--- a/app/models/teacher_training_adviser/steps/identity.rb
+++ b/app/models/teacher_training_adviser/steps/identity.rb
@@ -6,6 +6,7 @@ module TeacherTrainingAdviser::Steps
     attribute :last_name, :string
     attribute :email, :string
     attribute :channel_id, :integer
+    attribute :sub_channel_id
 
     validates :first_name, presence: true, length: { maximum: 256 }
     validates :last_name, presence: true, length: { maximum: 256 }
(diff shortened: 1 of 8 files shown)


gemmadallmandfe:
@ethax-ross The error message makes more sense now :-)

What I don't quite understand is why we need the notification banner when clicking on the error message link effectively resets the form as well? Does it not clear the channel id?

gemmadallmandfe:
I still don't quite understand this @ethax-ross 

Clicking on the two links takes you to two different screens? Why do we need both?

image

ethax-ross:
@gemmadallmandfe oh I see -- so we can't change the link (at least easily) in the validation message so that's why we need the notification; but I see what you mean we may as well just not show the validation error -- I'll updated that.

peteryates (approved):
Looks good to me

gemmadallmandfe:
This looks fine to me as far as I can tell!

Tested as follows:

sub_channel - used a query string with a sub_channel and went through the TTA funnel to completion

channel - used a query string with an invalid channel and went through the TTA funnel to completion

PR #968, opened 2022-03-16
Upgrade libxslt and libxm2

ethax-ross:
There are vulnerabilities in the versions of these packages within the alpine-3.15 image; manually upgrading them to placate snyk.


peteryates (approved):
Looks good

PR #973, opened 2022-03-18
Run through full sign up as existing candidate

ethax-ross:
The existing sign up integration test is currently failing because the candidate is not in the required state to prevent them re-signing up.

Instead, change the test to do a full sign up using an existing candidate record, which is arguably more thorough/useful anyway.


peteryates (approved):
Looks good to me

PR #987, opened 2022-04-08
Migrate production credentials to Azure

ethax-ross:
### Trello card

[Trello-1660](https://trello.com/c/eRpJcktw/1660-migrate-all-secrets-from-rails-credentials-to-the-azure-keyvaults)

### Context

We want to store all credentials in Azure for better security.

- Remove RAILS_ENV from dotenv files

We need to keep these in the Azure secrets (or update the CI to inject them manually) as Rails has already booted by the time dotenv runs.

- Make userresearch inherit from test environment

Simplifies things and makes it a bit more explicit.

- Remove production Rails credentials

We have migrated all of these into Azure so no longer need the Rails credentials.


peteryates (approved):
👍🏽

config/environments/userresearch.rb
```diff
@@ -1,9 +1,6 @@
 # Just use the production settings
```
peteryates: Does this comment still make sense?

peteryates (approved):

PR #988, opened 2022-04-08
Migrate local development secrets to Azure

ethax-ross:
### Context

We want to store our local development secrets in Azure instead of in a Rails credentials file.

- Migrate local development secrets to Azure

Remove development Rails credentials (now added to a new `s146d01-local2-kv` keyvault).

Add Makefile command to pull the local development secrets from Azure into a `.env.development` dotenv file (which is now gitignored).

We can continue to hard-code non-secret environment variables into `.env.development.yml` and these will be merged with the Azure secrets on running the setup command `make local setup-local-env`.

Update README to reflect new setup instructions.


peteryates (approved):
Makes sense.

PR #991, opened 2022-04-11
Remove references to Rails credentials

ethax-ross:
### Trello card

[Trello-1660](https://trello.com/c/eRpJcktw/1660-migrate-all-secrets-from-rails-credentials-to-the-azure-keyvaults)

### Context

We no longer use Rails credentials but some of the code will default to the Rails creds if an ENV var is not set; we no longer need to do that.

- Remove references to Rails credentials


peteryates (approved):
Much tidier

PR #997, opened 2022-04-20
[546] Rebuild docker cache workflow

saliceti:
### Context
Docker build layers are cached to optimise build performance but sometimes the base image is updated so we need to rebuild it.

Rebuild once a week on Sunday or run manually as required.

### Guidance to review
Run the new workflow


saliceti:
[Add workflow to bust docker cache](https://trello.com/c/HwLOMQ9D/546-add-workflow-to-bust-docker-cache)

PR #1021, opened 2022-05-17
Ensure all pages have correct titles

ethax-ross:
### Trello card

[Trello-3248](https://trello.com/c/npC4lx2i/3248-dac-audit-gov-page-title)
[Trello-3249](https://trello.com/c/73iRG3Do/3249-dac-audit-gov-error-title)

### Context

Make sure all pages have an appropriate page title in the format:

`Main heading - Service name - GOV.UK`

If there is an error on the page then the title should be prefixed with `Error:`.

- Ensure all pages have a correct page title
- Correct typo in make command

### Guidance to review

Changes inline with [GDS guidance](http://hmrc.github.io/assets-frontend/components/page-title/index.html)

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@
 1. Run `bundle install` to install the gem dependencies
 2. Run `yarn` to install node dependencies
 3. Run `rails db:setup` to setup the database.
-4. Run `az login` and then `make local set-local-env` to populate development secrets
+4. Run `az login` and then `make local setup-local-env` to populate development secrets
 5. Run `bundle exec rails server` to launch the app on http://localhost:3000
 6. (optional) Run `./bin/webpack-dev-server` in a separate shell for faster compilation of assets
 
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -9,12 +9,10 @@ class PagesController < ApplicationController
   end
 
   def session_expired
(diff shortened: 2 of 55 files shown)


peteryates (approved):
Looks good overall.

From an accessibility point of view the Design System recommends prefixing pages with an error on them to "Error: regular title"

> As well as showing an error summary, follow the [validation pattern](https://design-system.service.gov.uk/patterns/validation/) - for example, by adding ‘Error: ’ to the beginning of the page <title> so screen readers read it out as soon as possible.

If we were to wrap this setting of titles in a helper, could we pass in both the object (if there is one) and the title, so it could add a prefix accordingly? 🤔

```erb
<% set_page_title("Cookies on Get into Teaching", object: @mailing_list_signup) %>
```

ethax-ross:
@peteryates that's already in the PR, albeit in a slightly different format?

https://github.com/DFE-Digital/get-teacher-training-adviser-service/pull/1021/files#diff-5f3fc5e3977d242572aa1d08551f5eb557de0ccaff30370838ee9df5386ea0daR13

```
def format_page_title(title, status)
  prefix = status == 422 ? "Error: " : ""
  title = [title, "Get an adviser", "GOV.UK"].compact.join(" - ")

  "#{prefix}#{title}"
end
```

PR #1022, opened 2022-05-18
Skip steps the candidate has seen/submitted already

ethax-ross:
### Trello card

[Trello-3250](https://trello.com/c/lAuxvW5N/3250-dac-audit-gov-check-answers)

### Context

At the moment the candidate will have to go through the remainder of the wizard steps when they change an answer from the check answers page.

Instead, we want to jump them straight back to the check answers page if they have already completed the form.

- Skip steps the candidate has seen/submitted already

Pull in updated `dfe_wizard` gem which has this behaviour. Ensure the check answers step is always displayed to the user by overriding `#seen?`.

Default country to UK during data export instead of within the `uk_or_overseas` step; if we do it in the step it ends up setting the `country_id` and so the `overseas_country` step will be skipped if pre-filled when it shouldn't be (as its not optional).

### Guidance to review

[Corresponding dfe_wizard PR](https://github.com/DFE-Digital/dfe_wizard/pull/7)

Things to check:

- The various TTA wizard flows still work
- Pre-filled optional steps are still skipped
- Exit steps (e.g. no degree) still behave
- Check answers and change works inline with the GDS guidance
- Check answers then change to get a new flow works correctly

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/DFE-Digital/dfe_wizard.git
-  revision: 68278550c0a4771300d697e845ed213e58921009
+  revision: ba7ec290f046c9abb7361a159ff296afa053f69b
   specs:
     dfe_wizard (1.0.0)
       activemodel (>= 6.0.3.4)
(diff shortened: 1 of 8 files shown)


peteryates (approved):
Looks good.

Dynamics is leaking again though, countries already have a primary key 😢

gemmadallmandfe:
@ethax-ross This looks pretty good to me - it's a much better experience

One question:

**Name/Email**
If you change your name or email, you have to go through the whole funnel again
I'm presuming this is deliberate as it could be an entirely different person, but wanted to check!

It'll be interesting to see if this has any impact on completion rate @zali88

ethax-ross:
@gemmadallmandfe yeah that's deliberate; as we matchback and potentially pre-fill subsequent steps we clear the journey (its always been that way)

gemmadallmandfe (approved):
Looks good to me

PR #1023, opened 2022-05-18
Fix skip to main content behaviour

ethax-ross:
### Trello card

[Trello-3259](https://trello.com/c/FIlbKkHq/3259-dac-audit-skip-links)

### Context

Currently the skip to main content link appears below the cookie banner. This means its not the first thing you get to when you hit tab after the page loads and it also means the focus goes to the accept cookies button after selecting the skip to main content link (when it should go to the first focusable item in the main content).

Move the cookie banner to be below the skip to main content link so that it doesn't steel the initial focus. Update the cookie banner JS to stop the auto-focus onto the accept cookies button (I'm not sure why we were doing this in the first place).

- Fix skip to main content behaviour


peteryates (approved):
👍🏽

PR #1024, opened 2022-05-18
Remove vertical bars from footer links

ethax-ross:
### Trello card

[Trello-3255](https://trello.com/c/BTwLb2dR/3255-dac-audit-vertical-bar-in-links)

### Context

The vertical bars are characters and end up being read out by screen readers; we should get rid of them as they don't add any value.

Clean up the CSS to use the GOV.UK footer classes.

- Remove vertical bars from footer links


peteryates (approved):
Looks good

PR #1026, opened 2022-05-19
Ensure all pages have a main landmark

ethax-ross:
### Trello card

[Trello-3257](https://trello.com/c/twvsWl1m/3257-dac-audit-no-main-landmark)

### Context

The completed page did not contain a main landmark and so the skip to main content link was not working.

Update application layout to include the main landmark and add a new `yield` so that we can insert back links above it on the form steps.

- Ensure all pages have a main landmark
- Patch freetype vulnerability

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -57,7 +57,13 @@
       </div>
     </div>
 
-    <%= yield %>
+    <% if content_for?(:above_main_content) %>
+      <%= yield(:above_main_content) %>
+    <% end %>
+
+    <main class="govuk-main-wrapper no-margin" id="main-content" role="main">
+      <%= yield %>
+    </main>
 
     <%= render "layouts/footer" %>
   <% end %>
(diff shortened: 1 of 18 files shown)


app/views/layouts/application.html.erb
```diff
@@ -57,7 +57,13 @@
       </div>
     </div>
 
-    <%= yield %>
+    <% if content_for?(:above_main_content) %>
+      <%= yield(:above_main_content) %>
+    <% end %>
+
+    <main class="govuk-main-wrapper no-margin" id="main-content" role="main">
```
peteryates, line 64: This makes so much more sense 😅

gemmadallmandfe (approved):
Looks good to me too - happy for this to be merged

Let me know if we need a separate ticket re a check for main on all pages

ethax-ross:
@peteryates I thought about something like that but decided against it as we only have the one layout in the TTA service and now it contains the `main` element; if we added a new layout the spec would probably miss it anyway 😞

PR #1027, opened 2022-05-19
Prevent SVG logos from being read out by screen readers

ethax-ross:
### Trello card

[Trello-3258](https://trello.com/c/tK1s74V5/3258-dac-audit-gov-logo)

### Context

The GOV.UK and OGL logos are SVGs with a `role="presentation"`. They should instead be hidden from screen readers with `aria-hidden="true"`. The GOV.UK logo in particular results in an unhelpful voice-over from screen readers of 'slash' as they read out the `href` of the surrounding anchor.

- Prevent SVG logos from being read out by screen readers


peteryates (approved):
Looks good.

This change was also made in [Design System 4.0.0](https://github.com/alphagov/govuk-frontend/releases/tag/v4.0.0) - (/alphagov/govuk-frontend/pull/1724)

PR #1033, opened 2022-06-07
Upgrade to Ruby 3.1.2

ethax-ross:
### Trello card

[Trello-3345](https://trello.com/c/I9CiRkce/3345-upgrade-ruby-rails-in-tta-service)

### Context

Upgrade to the latest Ruby version ahead of upgrading to Rails 7.

- Upgrade to Ruby 3.1.2

Upgrade to the latest Ruby version ahead of upgrading to Rails 7. Ruby 3.1 moved `net-smtp` to be a bundled gem and whilst we don't send emails from the app Rails appears to still reference it.

- Update validates_timeliness gem

The current release (v5) doesn't adhere to the new splat operator requirement for hash -> named arguments, so we need to pull in a pre-release until they cut a new version.

- Use splat operator on hash -> named arguments

Ruby 3.1 requires an explicit splat operator when passing a hash in for named arguments.

On the form builder use cases we can simply omit the empty hash.

- Fix Rubocop linting errors

We can omit the hash value when the key matches the variable name referenced.

- Replace deprecated throttled_response method call

The replacement is called throttled_responder so we're using that instead

https://github.com/rack/rack-attack/blob/master/CHANGELOG.md


peteryates (approved):
Nice!

PR #1034, opened 2022-06-08
Upgrade to Rails 7.0.3

ethax-ross:
### Trello card

[Trello-3345](https://trello.com/c/I9CiRkce/3345-upgrade-ruby-rails-in-tta-service)

### Context

We want to run on the latest version of Rails.

- Upgrade to Rails 7

Upgrade Rails to v7.0.3 and run the `rails app:update` rake task bringing latest defaults into initializers.

Remove old framework defaults and go straight to the v7.0 defaults.

### Guidance to review

I could have had it so the environment files were easier to diff but at least this way when we need to run `rails app:update` again it should be more straight forward.


peteryates (approved):
All seems sensible

PR #1045, opened 2022-06-22
Fix missing translation for time zone on check answers page

ethax-ross:
### Trello card

[Trello-3416](https://trello.com/c/kx8angpq/3416-fix-time-zone-label-error-on-tta-funnel-check-answers-page)

### Context

There is a missing translation for the 'time zone' row on the check answers page, causing it to render as 'Title'.

- Add missing translation for overseas time zone

The translation was missing for the overseas time zone field, causing it to display 'Title' on the check answers page instead of 'Time zone'. It looks like it was placed under the wrong key in the translations file.

- Use in-memory store in development

If the Rails cache is disabled we were defaulting to a `null_store` for the cache; we later set the `session_store` to use the `cache_store`, however, so it was causing the session to immediately expire. Instead, we should use a `memory_store` in development (caching is disabled by setting `perform_caching` to false).

I'm not sure how this ever worked 😕

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -17,20 +17,20 @@ Rails.application.configure do
   # Enable server timing
   config.server_timing = true
 
+  # Use in-memory cache
+  config.cache_store = :memory_store
+
   # Enable/disable caching. By default caching is disabled.
   # Run rails dev:cache to toggle caching.
   if Rails.root.join("tmp/caching-dev.txt").exist?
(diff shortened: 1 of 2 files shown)


gemmadallmandfe (approved):
Looks good to me - not sure if you need @peteryates 's tech eyes on it though

peteryates (approved):
Looks good

PR #1054, opened 2022-07-01
Fix 302 redirect on root sign up path

ethax-ross:
The `dfe_wizard` gem has been updated to ensure the root redirect from `/signup` to `/signup/identity` is a 301 (permanent) instead of 302 (temporary).


gemmadallmandfe (approved):
Looks good to me - tested redirect and it is now a 301

gemmadallmandfe (approved):

PR #1049, opened 2022-06-27
updating RTTA eligibility

EllieNodder:
### Trello card

https://trello.com/c/Prol5Z8w/3397-remove-subject-eligibility-for-rttas

### Context

Now that RTTA eligiblity has expanded, we need to update the adviser funnel.

PR #1055, opened 2022-07-01
Allow returning teachers to specify any secondary subject preference

ethax-ross:
> :warning: Pending updated copy for the new exit step and potentially the primary/secondary step as well when used in the returners journey

### Trello card

[Trello-3397](https://trello.com/c/Prol5Z8w/3397-remove-subject-eligibility-for-rttas)

### Context

We currently only accept certain secondary subjects on the returning teacher journey - if the candidate does not want to teach these subjects and selects "Other" we block their sign up and let them know they are not eligible.

Going forward we are opening up returning teacher submissions for any secondary subject, so we want to include all subjects (except "Primary") in the list of preferred subjects the candidate would like to teach and add a new step that checks if they want to teach primary or secondary. If they choose secondary we will block their sign up and let them know they are not eligible.

- Include StageInterestedTeaching step in returning teacher journey
- Include all secondary subjects in SubjectLikeToTeach step
- Move education stage exit step to be after StageInterestedTeaching step

### Guidance to review

Other candidate sign up journeys _shouldn't_ be effected by these changes, but it's worth doing a bit of exploratory testing.

Best reviewed by-commit.


EllieNodder:
@ethax-ross I made a separate [PR](https://github.com/DFE-Digital/get-teacher-training-adviser-service/pull/1049) for updating the funnel start page - can we merge these two together?

app/views/teacher_training_adviser/steps/_primary_returner.erb
```diff
@@ -5,7 +5,7 @@
     <div class="govuk-grid-column-two-thirds">
       <h1 class="govuk-heading-l">We're sorry, but you are not eligible for this service.</h1>
 
-      <p>To be eligible, you must be planning to teach chemistry, computing, maths, modern foreign languages or physics.</p>
+      <p>To be eligible, you must be planning to teach a secondary subject.</p>
```
EllieNodder, line 8:
```suggestion
      <p>If you already have qualified teacher status (QTS) and are returning to teaching, you must be planning to teach a secondary subject to be eligible.</p>
```
@gemmadallmandfe do we think it might be good to be clearer here? Just in case someone has accidentally navigated to this end screen, then thinks **all** candidates need to be teaching a secondary subject to get an adviser, not just returning teachers.

ethax-ross:
@EllieNodder @gemmadallmandfe I've pulled Ellie's content changes into this PR and also implemented the suggested update to the exit step content.

app/views/teacher_training_adviser/steps/_primary_returner.erb
gemmadallmandfe: @EllieNodder Yes, I think that would make sense

Maybe '.... to be eligible for an adviser.'? The page is a bit odd as it stands as it talks about being eligible for the service but doesn't really say what it is - is that deliberate?

gemmadallmandfe:
@ethax-ross @EllieNodder 

Polly has flagged a small issue with the funnel - the help text on the 'Select a stage' step does not make sense; this *does* have an impact on sign-up as you are only eligible if you are choosing secondary:

image


@ethax-ross Is it possible to suppress this help text but only for returners?

@EllieNodder Should we include different help text or is this OK being removed altogether?

gemmadallmandfe:
@ethax-ross Thanks for changing that - looks good to me. Help text is still present for the ITT route.

Will check with @EllieNodder that it makes sense from a content perspective when she is back next week

EllieNodder:
> @ethax-ross Thanks for changing that - looks good to me. Help text is still present for the ITT route.
> 
> Will check with @EllieNodder that it makes sense from a content perspective when she is back next week

Looks good to me

gemmadallmandfe:
The returners policy team have approved the changes and given the go-ahead for this to be merged to production

gemmadallmandfe (approved):
Looks good to me

gemmadallmandfe:
@leoapost Do you need to have tech eyes on this or can it just be merged?

PR #1091, opened 2022-08-17
Add dfe-analytics gem

ethax-ross:
### Trello card

[Trello-3507](https://trello.com/c/3CzgfQK9/3507-add-dfe-analytics-gem-to-services)

### Context

Add the `dfe-analytics` gem and enable it when running in non-production environments. Capture all web requests and model updates as events (we only have a single `Feedback` model we are capturing).

Minor tweak to the `request_id` specs to allow `request.uuid` to be called multiple times; it appears the `dfe-analytics` gem also requests the `uuid` so we need to allow the mock to be called more than once.

The JSON secret for BigQuery has been added to the Azure keyvault; I had to base64 encode it otherwise the private key was mangled.

- Add dfe-analytics gem

### Guidance to review

The TTA app doesn't use `ActiveJob` so it's currently configured with the default Async (in memory) queue adapter. This means jobs may be lost on restart. We could setup Sidekiq backed by our existing Redis instance as an improvement if we think that would be worthwhile.


misaka (approved):
LGTM!!

PR #1118, opened 2022-09-06
Updating link back to GIT

EllieNodder:
### Context

Kat pointed out that the current wording could be confusing to returners, so I'm editing the copy so it applies to all users.


gemmadallmandfe (approved):
Looks good to me

PR #1160, opened 2022-10-24
Update various dependencies

ethax-ross:
### Trello card

[Trello-3808](https://trello.com/c/xYgwOwq5/3808-fix-csp-errors-on-tta-service)

### Context

The TTA service has various dependencies with major version bumps that we need to update and require a bit of refactoring for compatibility. 

- Upgrade to Stimulus v3

Upgrade to the latest version of Stimulus.

- Upgrade Redis to v5

Using a single connection to Redis is now discouraged because it can lead to blocking. The recommended fix is to enable connection pooling.

Wrap the existing connection using the connection_pool gem.

- Update dfe-analytics gem

Update to the latest version of `dfe-analytics`. There is an issue whereby the latest gem calls `eager_load!` which causes Rails to evaluate the `inclusion` validation block in `WhatDegreeClass` and make an external request (which fails due to being unauthorized in the test environment). It probably shouldn't do this if the gem is disabled in that environment.

### Guidance to review

I'll create a separate ticket to clean up the CSP directives as we should make them consistent with the GiT website for ease.

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -1,3 +1,6 @@
 VCAP_APPLICATION="{\"application_name\": \"app-name\",\"space_name\": \"space-name\",\"organization_name\": \"org-name\"}"
 GTM_ID=123-ABC
 CF_INSTANCE_INDEX=app-instance
+# Prevent the dfe-analytics gem eager-loading Rails (which
+# causes our validators to fire network requests).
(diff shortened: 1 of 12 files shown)


asatwal (approved):
👍  LGTM

PR #1161, opened 2022-10-25
Update CSP to whitelist domains

ethax-ross:
I'm not sure why these have started suddenly flagging as violations (the pixels have been enabled for months!).

Whitelist google ad services, reddit, hotjar and our GTM server-side container.

I plan on doing a clean up of the CSP directives in the future to get them more inline with those in the app and just generally tidied, but this fixes the immediate issues.


asatwal (approved):
👍  LGTM

PR #1163, opened 2022-10-26
Fix Privacy Policy formatting

ethax-ross:
Our HTML whitelisting was too strict and stripping out headings/jump links. I'm updating it to match the same method on the Get into Teaching website so that its consistent.


gemmadallmandfe (approved):
Looks good to me

PR #1164, opened 2022-10-27
Fix Hotjar CSP error

ethax-ross:
We need the HotJar domain in the `frame_src` directive as we see a violation in production just now.


asatwal (approved):
👍

PR #1179, opened 2022-11-08
Update cookie hide link text

ethax-ross:
### Trello card

[Trello-3847](https://trello.com/c/2CgyrH26/3847-dac-audit-non-descriptive-links-tta-cookies)

### Context

Our cookie acceptance message has a link with the text `hide` which is not very accessible for screen readers. Use `hide cookie message` instead.

- Update cookie hide link text


gemmadallmandfe (approved):
Looks good to me - tested on desktop and also on iPhone using VoiceOver

PR #1189, opened 2022-11-30
Remove privacy policy step from sign up journey

ethax-ross:
### Trello card

[Trello-3908](https://trello.com/c/zVEDgHSA/3908-update-tta-funnel-to-have-privacy-notice-link-prior-to-data-entry)

### Context

We have a link and text to call out the privacy notice on the first page of the sign up journey; there is not a requirement to have the user explicitly accept the privacy policy and so we can remove that step from the journey.

We still need to send the API/CRM the ID of the privacy notice we surfaced to the user during sign up, however, so we now capture this on the first step.

- Remove privacy policy step from sign up journey


gemmadallmandfe:
@ethax-ross Please could you update the copy on `/teacher_training_adviser/sign_up/identity` so it is consistent with the GIT funnels and also references "advisers" rather than "teacher training advisers" (so we cover RTTAs and ETAs)

* Our privacy notice explains how we use your personal data. It's important you read this before signing up to get support from an adviser.

cc @GayeCord for info

gemmadallmandfe:
@ethax-ross The privacy notice opens up in the same tab in the adviser funnel, but in a new tab in the mailing list, events and callback funnels on the GIT side

We should be consistent - which one is more appropriate in this case?

cc @GayeCord

gemmadallmandfe:
@ethax-ross The button at on `/teacher_training_adviser/sign_up/review_answers` says `Complete`

Can we change this to `Complete sign up` to be consistent with GIT funnels and to give more context which is better in terms of accessibility?

cc @GayeCord for info

gemmadallmandfe (approved):
Looks good to me

app/models/teacher_training_adviser/steps/identity.rb
```diff
@@ -22,12 +24,16 @@ def export
...
+      @latest_privacy_policy ||= GetIntoTeachingApiClient::PrivacyPoliciesApi.new.get_latest_privacy_policy
+    end
+
     def reviewable_answers
       super
         .tap { |answers|
           answers["name"] = "#{answers['first_name']} #{answers['last_name']}"
         }
-        .without("first_name", "last_name", "channel_id", "sub_channel_id")
+        .without("first_name", "last_name", "channel_id", "sub_channel_id", "accepted_policy_id")
```
asatwal, line 36: Nitpick: Might be clearer to use shorthand here:
``` Ruby 
 %w[first_name last_name channel_id sub_channel_id accepted_policy_id]
```

asatwal (approved):
👍 LGTM - Minor comment

app/models/teacher_training_adviser/steps/identity.rb
ethax-ross: I'll merge this as-is for now as I want a full day of it being out to keep an eye on it (as its Friday) but I'll make that change when I'm back in this model, which I have a ticket around DRYing up in general so should be very shortly.

PR #1211, opened 2022-12-15
Added warning of slower response during holidays

philspray:
### Trello card

https://trello.com/c/C8nLNnH0

### Context

As someone interested in teaching who has signed up for an adviser
I need to understand that there may be a delayed response over Christmas
So that my expectations are set correctly

Added a message to the completion page to explain there may be a delay in response times until January.


gemmadallmandfe (approved):
Looks good to me

PR #1215, opened 2022-12-21
Gracefully handle when there are no available callback slots

ethax-ross:
### Trello card

[Trello-4062](https://trello.com/c/n5iRbaIL/4062-review-callback-quota-fallback-functionality-in-light-of-increased-demand)

### Context

Currently the API will provide a fallback list of available callback slots in the event the CRM returns none. Going forward the API will also return 0 if the CRM has none available.

We need to account for this in the front-end by asking the candidates to contact us instead of providing a mechanism for them to book a callback.

- Gracefully handle when there are no available callback slots

Update copy on steps that ask for a callback; on the overseas journey its actually two steps as we ask for their phone number/time zone before showing a list of callback slots (in their selected time zone). If the candidate is unable to book a callback we will not ask for their telephone number/time zone.

### Guidance to review

There's currently no way to easily test this, but it should just present the same behaviour as master/production until I make the API change to allow 0 callback slots to be returned.

### DIFF
--- a/app/controllers/teacher_training_adviser/steps_controller.rb
+++ b/app/controllers/teacher_training_adviser/steps_controller.rb
@@ -10,7 +10,8 @@ module TeacherTrainingAdviser
       super
 
       @returner = wizard_store[:type_id].to_i == Steps::ReturningTeacher::OPTIONS[:returning_to_teaching]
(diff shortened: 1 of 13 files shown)


asatwal (approved):
👍 LGTM

PR #1228, opened 2023-01-19
Removing 'free of charge' from footer

EllieNodder:
### Trello card

https://trello.com/c/5wC3G0jC/4124-remove-free-of-charge-from-adviser-funnel-footer-re-contact-centre-calls

### Context

The adviser funnel footer currently says the phone line is free of charge, but this isn't true for international users, so we should remove this.


gemmadallmandfe:
Looks fine @EllieNodder :-) 

My only question is whether we can have a bit of a gap before the OGL line as we have currently, so the contact details don't get lost? Might be a question for @ethax-ross 

image

philspray (dismissed):
Looks good to me.

gemmadallmandfe (approved):
Looks good to me

PR #1232, opened 2023-01-26
Remove address line 1/2 and city

ethax-ross:
### Trello card

[Trello-4135](https://trello.com/c/J3a7zgtl/4135-remove-the-address-fields-from-the-adviser-funnel)

### Context

On the UK address step we only want to ask candidates for their postcode (we currently ask for their full address).

- Remove address line 1/2 and city.


gemmadallmandfe:
@ethax-ross Tested with Tejas - record captured successfully in CRM with just postcode

One minor issue - on the completion page, the post code is listed against Title - the label is wrong?

image

app/views/teacher_training_adviser/steps/_uk_address.html.erb
```diff
@@ -1,8 +1,5 @@
-<% @page_title = "What is your address?" %>
+<% @page_title = "What is your postcode?" %>
 
-<%= f.govuk_fieldset legend: { text: 'What is your address?' } do %>
-  <%= f.govuk_text_field :address_line1, width: 20 %>
-  <%= f.govuk_text_field :address_line2, width: 20 %>
-  <%= f.govuk_text_field :address_city, width: 20 %>
-  <%= f.govuk_text_field :address_postcode, width: 20 %>
+<%= f.govuk_text_field :address_postcode, width: 20, label: { text: @page_title, size: "m", tag: "h1" } do %>
+  <p>We try and match you with an adviser providing support in your region, or based on the subject you’re interested in teaching.</p>
```
gemmadallmandfe, line 4:
```suggestion
  <p>We try and match you with an adviser providing support in your region.</p>
```
@EllieNodder Does talking about subject here sound a bit confusing? Is it sufficient to leave as just talking about region, given that we say 'try'?

gemmadallmandfe: NB This screen also comes up in the returners flow

EllieNodder: @gemmadallmandfe wasn't sure if we needed to add the subject bit, given that 'We try and and match you with an adviser providing support in your region' won't always be true if, for ETAs, we're actually mostly trying to match them by subject?  Let's discuss with Andy and am happy to change if you both think we should get rid of the subject bit?

gemmadallmandfe:
Tested end to end with Tejas for an ETA, TTA and RTTA

All records had a region assigned correctly based on postcode

Ready to deploy when Andy H gives the go-ahead from a wording perspective @EllieNodder

asatwal (approved):
👍  LGTM

PR #1236, opened 2023-01-30
Upgrade optimize-css-assets-webpack-plugin

ethax-ross:
### Trello card

[Trello-4160](https://trello.com/c/tqXsfBuY/4160-fix-snyk-vulnerability)

### Context

There is a vulnerability in `nth-check`, which is a distant dependency of `webpacker` by way of `optimize-css-assets-webpack-plugin`. Manually upgrading this plugin to patch the security issue.

- Upgrade optimize-css-assets-webpack-plugin

### Guidance to review

We probably want to look at upgrading `webpacker` to one of the newer Rails 7 assets pipelines in the near future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,8 @@
     "serialize-javascript": "^6.0.0",
     "set-value": "^4.0.1",
     "stimulus": "^3.2",
-    "turbolinks": "^5.2.0"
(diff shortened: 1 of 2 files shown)


asatwal (approved):
👍  LGTM

PR #1250, opened 2023-02-17
Migrate to Shakapacker

ethax-ross:
### Trello card

[Trello-4245](https://trello.com/c/3viAxYsH/4245-remove-webpacker-from-git-and-get-an-adviser-apps)

### Context

Webpacker is no longer maintained and several of the dependencies now have vulnerabilities that, whilst patched, haven't made it upstream into the Webpacker gem and likely won't.

- Migrate to Shakapacker

Migrate to Shackapacker; whilst there are other build tools that we can use for Rails 7 this is the path of least resistance and brings the application inline/consistent with other GiT apps.

Notable changes;

- Shackapacker only allows a single `javascript_pack_tag` (which supports multiple JS files as arguments). This required a minor update to the GTM pack to conditionally boot.
- Binstubs updated and are now `webpacker*` rather than `webpack*`.
- Paths to static files had to be updated to match the new manifest.json
- Exclude IE11 from the browesrlist

It may be worth switching to `esbuild` in the future but for now I've retained `babel` for transpiling.

### Guidance to review

Worth a general browse of the site with the console open to check for any JS errors or issues including CSS files. Also worth testing GTM specifically.


asatwal (approved):
👍  LGTM

PR #1253, opened 2023-02-22
Changing subject interested in teaching copy

EllieNodder:
### Trello card

https://trello.com/c/8kSTY8rg/4253-review-teaching-subject-question-for-eta-candidates-in-adviser-funnel

### Context

The current 'What subject would you like to teach?' question doesn't really work for those interested in teaching primary, now that we've taken the 'What stage do you want to teach?' question out for ETAs.

We should tweak the question and add hint text to make the question work for ETAs.


ethax-ross:
Changes pulled into #1271

PR #1271, opened 2023-03-07
Re-introduce education stage step for ETAs

ethax-ross:
### Trello card

[Trello-4253](https://trello.com/c/8kSTY8rg/4253-review-teaching-subject-question-for-eta-candidates-in-adviser-funnel)

### Context

We removed the education stage step for ETAs but this led to a confusing subject you'd like to teach step where they had to select 'primary' as their subject if they wanted to teach at primary stage.

- Re-introduce education stage step for ETAs

Re-introduce the education stage step for all candidates. Remove 'primary' from the subject you'd like to teach step as an option (as we only display this step to those interested in teaching secondary). Update the helper text for the subject you'd like to teach step.


EllieNodder (commented):
Looks good to me!

gemmadallmandfe:
@EllieNodder Whilst we are addressing this stage step, do you think we should improve the hint text
eg to say that you can change your mind later? and maybe that it helps us match them with an adviser?

(What we are saying is not strictly true - you get asked slightly different questions depending on  your answer so it does affect the sign-up process to that extent!)

image

gemmadallmandfe:
Tested with Tejas to ensure that the requests were still being allocated to the right dashboards in CRM - all looks good

Once the hint text has been changed, I'm happy to approve/merge this

gemmadallmandfe:
Candidate experience team are happy for this to be put into prod

I'll approve/merge on Monday so we can monitor

gemmadallmandfe (approved):
Looks good to me

PR #1277, opened 2023-03-16
Update accessibility statement with holding text

gemmadallmandfe:
### Trello card
https://trello.com/c/P0njxMzH

### Context
We want to update our accessibility statement to say that we have booked our next external audit.


GayeCord:
looks great :-)

PR #1317, opened 2023-05-19
Add Google Optimize for split testing

ethax-ross:
> ⚠️ Not to be merged until we are ready to run the split test

### Trello card

[Trello-4574](https://trello.com/c/ftxK1h2x/4574-implement-google-optimize-on-get-an-adviser)

### Context

We are split testing the TTA sign up on the GiT website with the Get an adviser service; in order for the test to work correctly the optimize script needs to be on both variants. It's already on the GiT website, but this will pull it into the Get an adviser service home page.

- Add Google Optimize for split testing

### Guidance to review

This is largely a lift and shift of the implementation from the GiT website.


ethax-ross:
As most people don't opt-in to cookies on the TTA service it will skew the optimize results, so we're only going to run it on the  GiT website for splitting the traffic

PR #1372, opened 2023-08-22
Various dependency updates

plisovin:
### Context

Just trying to get rid of PRs from dependabot to reduce used space


asatwal (approved):
👍  LGTM

